PR #2, opened 2020-09-09
Swap 1018 add comment input on sample review form

jekabs-karklins:


.eslintrc
```diff
@@ -51,6 +51,7 @@
       }
     ],
     "@typescript-eslint/no-unused-vars": "warn",
+    "@typescript-eslint/ban-ts-ignore": "warn",
```
martin-trajanovski, line 54: @jekabs-karklins Do we really want this to ignore typescript compilation errors sometimes? Or maybe this is just temporary.

jekabs-karklins: Yes, it's temporary, until they fix the graphql ISettings types.

martin-trajanovski (approved):

PR #7, opened 2020-09-14
smaller docker image

bolmsten:


jekabs-karklins (approved):
Looks good

martin-trajanovski:
I think this is already in place with one of the previous PRs so we don't need to merge this one.

jekabs-karklins:
Already on the develop

PR #8, opened 2020-09-14
Swap 1042 dev environment

peter-asztalos:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:12-slim
```
jekabs-karklins, line 1: Conflicts https://github.com/UserOfficeProject/user-office-backend/pull/7/files, but should go without any conflict, because the same

jekabs-karklins (approved):

Dockerfile
peter-asztalos: Oh, didn't see that PR, I'll resolve the conflict later when the other gets merged.

martin-trajanovski (approved):

PR #10, opened 2020-09-17
SWAP-1044 Scheduled event crud

peter-asztalos:
- Made User type to be accessible from other services
- Apollo federation fixes

### DIFF
--- a/src/middlewares/graphql.ts
+++ b/src/middlewares/graphql.ts
@@ -6,6 +6,7 @@ import baseContext from '../buildContext';
 import { ResolverContext } from '../context';
 import { Role } from '../models/Role';
 import { User, UserWithRole } from '../models/User';
+import federationSources from '../resolvers/federationSources';
 import { registerEnums } from '../resolvers/registerEnums';
 import { buildFederatedSchema } from '../utils/buildFederatedSchema';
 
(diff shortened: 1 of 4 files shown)


src/queries/UserQueries.ts
```diff
@@ -23,6 +23,10 @@ export default class UserQueries {
     return this.dataSource.get(id);
   }
 
+  async byRef(id: number) {
```
martin-trajanovski, line 26: @peter-asztalos isn't this the same as `getAgent` function that already exists here or this is used just for an example?

src/middlewares/graphql.ts
```diff
@@ -22,20 +23,29 @@ const apolloServer = async (app: Express) => {
...
+    }
+  );
 
   const server = new ApolloServer({
     schema,
     tracing: false,
     playground: {
       settings: {
+        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
         // @ts-ignore-line igore until https://github.com/prisma-labs/graphql-playground/pull/1212 is merged
```
martin-trajanovski, line 49: @peter-asztalos I think this https://github.com/graphql/graphql-playground/pull/1212 is already merged. Maybe we can remove the `ts-ignore`

martin-trajanovski (approved):

src/queries/UserQueries.ts
peter-asztalos: Yeah, I was thinking a bit about this, the `getAgent` has a bit misleading name and it might change in the future (auth requirements e.g. ?) so I wanted to have a different method for the sole purpose for getting a user referenced by other services, that's why it's called `byRef` so it suggest it is used to fetch a user by some reference.

src/middlewares/graphql.ts
peter-asztalos: We use `ApolloServer` and they still have the old version.

jekabs-karklins (approved):

PR #11, opened 2020-09-18
Swap 1050 proposal workflow

martin-trajanovski:
This one closes SWAP-1056(use of proposal statuses new structure) and adds basic parts of proposal workflows. Another one is coming to close the SWAP-1050.


src/resolvers/mutations/settings/CreateProposalWorkflowMutation.ts
```diff
@@ -0,0 +1,31 @@
+import { Args, ArgsType, Ctx, Mutation, Resolver, Field } from 'type-graphql';
+
+import { ResolverContext } from '../../../context';
+import { ProposalWorkflowResponseWrap } from '../../types/CommonWrappers';
+import { wrapResponse } from '../../wrapResponse';
+
+@ArgsType()
```
peter-asztalos, line 7: I have seen this in the code base a few times, but instead of `ArgsType` shouldn't we use `InputType` for mutations, as `Arg` and `ArgsType` is for queries?

peter-asztalos (approved):

src/resolvers/mutations/settings/CreateProposalWorkflowMutation.ts
martin-trajanovski: I am going to fix it now for the new mutations added in this PR. And later we can start fixing all the others.

PR #18, opened 2020-09-24
Swap 1050 proposal workflow

martin-trajanovski:


db_patches/0060_AddProposalWorkflows.sql
```diff
@@ -17,6 +17,7 @@ BEGIN
             proposal_status_id INT REFERENCES proposal_statuses (proposal_status_id) ON DELETE CASCADE NOT NULL,
             next_proposal_status_id INT REFERENCES proposal_statuses (proposal_status_id) ON DELETE CASCADE,
             prev_proposal_status_id INT REFERENCES proposal_statuses (proposal_status_id) ON DELETE CASCADE,
+            sort_order INTEGER NOT NULL,
```
peter-asztalos, line 20: If the current deployed version already ran this patch before then it won't pick up this change, you may need an additional patch file for this

peter-asztalos (approved):

db_patches/0060_AddProposalWorkflows.sql
martin-trajanovski: Yes I was thinking about this maybe it is better to put it in separate patch. 👍

PR #19, opened 2020-09-28
Scheduler e2e testing

peter-asztalos:


docker-compose.e2e.yml
```diff
@@ -0,0 +1,39 @@
...
+      DATABASE_URL: postgres://duouser:duopassword@db:5432/duo
+      secret: qMyLZALzs229ybdQXNyzYRdju7X784TH
+      tokenLife: 7d
+      baseURL: localhost:3000
+      SPARKPOST_TOKEN: dummytoken
+      NODE_ENV: production
+    labels:
+      - "traefik.frontend.rule=PathPrefix:/api, /graphql, /downloads, /files, /proposal, /uploads"
+      - "traefik.port=4000"
+  duo-cron-job:
```
jekabs-karklins, line 17: We don't need duo-cron-job here I believe.

peter-asztalos: Good point, I just copy pasted from dev

jekabs-karklins (approved):

PR #21, opened 2020-09-28
Swap 1061 as a user i want to clone a sample

jekabs-karklins:


src/datasources/postgres/QuestionaryDataSource.ts
```diff
@@ -6,6 +6,7 @@ import {
   Questionary,
   QuestionaryStep,
 } from '../../models/Questionary';
+import { logger } from '../../utils/Logger';
```
peter-asztalos, line 9: I think we could start using the logger moved to the npm package

src/datasources/postgres/QuestionaryDataSource.ts
```diff
@@ -300,4 +301,39 @@ export default class PostgresQuestionaryDataSource
...
+      )
+      SELECT 
+          ${clonedQuestionary.questionaryId}
+        , question_id
+        , answer
+      FROM 
+        answers
+      WHERE
+          questionary_id=${sourceQuestionary.questionaryId}
+    `);
```
peter-asztalos, line 335: I think it would be more safe if we used the [parameter bindings](http://knexjs.org/#Raw-Bindings)

src/mutations/SampleMutations.ts
```diff
@@ -117,4 +119,29 @@ export default class SampleMutations {
...
+    try {
+      const sourceSample = await this.dataSource.getSample(sampleId);
+      const clonedQuestionary = await this.questionaryDataSource.clone(
+        sourceSample.questionaryId
+      );
+      const clonedSample = this.dataSource.create(
+        clonedQuestionary.questionaryId!,
+        `Copy of ${sourceSample.title}`,
+        sourceSample.creatorId
+      );
```
peter-asztalos, line 138: I'm wondering if it would make sense to do this in transaction.

src/mutations/SampleMutations.ts
```diff
@@ -117,4 +119,29 @@ export default class SampleMutations {
...
+    if (!sampleAuthorization.hasWriteRights(agent, sampleId)) {
+      return rejection('NOT_AUTHORIZED');
+    }
+
+    try {
+      const sourceSample = await this.dataSource.getSample(sampleId);
+      const clonedQuestionary = await this.questionaryDataSource.clone(
+        sourceSample.questionaryId
+      );
+      const clonedSample = this.dataSource.create(
```
peter-asztalos, line 134: `await` is missing from here

martin-trajanovski (approved):

src/datasources/postgres/QuestionaryDataSource.ts
jekabs-karklins: Good idea, but then let's do it in all places, so that it is consistent. I have added a task and we can take it in next sprint.
https://jira.esss.lu.se/browse/SWAP-1090

src/datasources/postgres/QuestionaryDataSource.ts
jekabs-karklins: I agree, but could not figure out how to do it in Knex elegantly

src/mutations/SampleMutations.ts
jekabs-karklins: I think it still works because the return type of function is Promise, but for readability I'll push the change

src/mutations/SampleMutations.ts
jekabs-karklins: Good idea, let me check

src/datasources/postgres/QuestionaryDataSource.ts
peter-asztalos: I think the named binding should look fine here, you can re-use it and define only once (hopefully)

src/mutations/SampleMutations.ts
peter-asztalos: Yes, but the try-catch wouldn't catch the error, it will be handled by an upper layer in the promise chain

PR #22, opened 2020-09-29
Swap 1081 proposal workflow validation

martin-trajanovski:


src/decorators/ValidateArgs.ts
```diff
@@ -33,6 +34,10 @@ const ValidateArgs = (schema: yup.ObjectSchema) => {
       const errors = await schemaValidation(schema, inputArgs);
 
       if (errors) {
+        logger.logError(`Input validation errors: ${errors}`, {
+          inputArgs,
+        });
```
peter-asztalos, line 39: I think we shouldn't log input values as they may contain sensitive information even if one of them is invalid. In the dev it may be helpful but then we should put it behind a environment check so it is not called in production.

martin-trajanovski: Yes that is true. I will add an environment check here it will be better to have it in development for debugging purpose.

jekabs-karklins (approved):

peter-asztalos (approved):

PR #27, opened 2020-10-07
SWAP-1085 Scheduled event by proposal

peter-asztalos:
Scheduler related changes


src/messageBroker.ts
```diff
@@ -1,23 +1,243 @@
...
+  properties: MessageProperties
+) => Promise<void>;
 
 export interface MessageBroker {
-  sendMessage(message: string): void;
+  sendMessage(queue: Queue, type: string, message: string): Promise<void>;
+  listenOn(queue: Queue, cb: ConsumerCallback): Promise<void>;
 }
 
 export class RabbitMQMessageBroker implements MessageBroker {
```
peter-asztalos, line 18: I'll move this to a separate package, but for now I'll leave this here.

peter-asztalos:
I want to move the message broker to a different package, but otherwise it is ready for review.

src/messageBroker.ts
peter-asztalos: Once https://github.com/UserOfficeProject/user-office-lib/pull/8 is merged I'll make the changes

src/decorators/EventBus.ts
```diff
@@ -37,7 +37,11 @@ const EventBusDecorator = (eventType: Event) => {
 
       // NOTE: Do not log the event in testing environment.
       if (process.env.NODE_ENV !== 'test') {
-        eventBus.publish(event);
+        eventBus
+          .publish(event)
+          .catch(e =>
+            console.error(`EventBus publish failed ${event.type}`, e)
```
jekabs-karklins, line 43: Maybe log an error instead, or will the error be logged elsewhere already?

src/eventHandlers/messageBroker.ts
```diff
@@ -0,0 +1,92 @@
...
+import { ProposalEndStatus } from '../models/Proposal';
+
+export default function createHandler({
+  reviewDataSource,
+  instrumentDataSource,
+}: {
+  reviewDataSource: ReviewDataSource;
+  instrumentDataSource: InstrumentDataSource;
+}) {
+  if (process.env.UO_FEATURE_DISABLE_MESSAGE_BROKER === '1') {
```
jekabs-karklins, line 16: If we use this naming convention UO_FEATURE_(DIS)_ENABLE_<FEATURE> than it would be nice to use it everywhere going forward

jekabs-karklins (approved):

src/eventHandlers/messageBroker.ts
peter-asztalos: Yes, that was my plan, or something similar, to be the same everywhere in the codebase.

src/queries/InstrumentQueries.ts
```diff
@@ -25,7 +25,9 @@ export default class InstrumentQueries {
     return instrument;
   }
 
-  @Authorized([Roles.USER_OFFICER])
+  // TODO: use some kind of shared API token between gateway and user-office to authorize the requests
+  // until some kind of auth gets implemented on the scheduler side
+  // @Authorized([Roles.USER_OFFICER])
```
mr-field, line 30: Is there an issue/plan to uncomment this in the future? It seems like a possible security problem if we forget about it

peter-asztalos: Yes, I'm already working on a task which will revert this change.

mr-field (approved):

PR #29, opened 2020-10-12
Swap 1052 attach workflow to a call

martin-trajanovski:


src/mutations/CallMutations.ts
```diff
@@ -101,4 +102,50 @@ export default class CallMutations {
         return rejection('INTERNAL_ERROR');
       });
   }
+
+  // @ValidateArgs(assignInstrumentsToCallValidationSchema)
```
jekabs-karklins, line 106: Is this waiting validation package push?

jekabs-karklins (approved):

src/mutations/CallMutations.ts
martin-trajanovski: Yes you are right I forgot to push there and enable this. Thanks 🥇

martin-trajanovski: Here is the pull request on the validation package https://github.com/UserOfficeProject/user-office-lib/pull/10

PR #33, opened 2020-10-20
Swap 1118 proposal workflow next status events

martin-trajanovski:
## Description

This pull request resolves the task: Ability to add multiple events that are triggering the change to the next status on a workflow connection.

## Motivation

Previously, we were able to add only one event that triggers change to the next status in the workflow. Now you can add multiple events and they will be combined with the `AND` operator later when we check for status change in the `WorkflowEngine`.

## Fixes:

* Task: Ability to add multiple events that are triggering the change to the next status on a workflow connection.

## Changes:

Backend adjustments for task resolution


jekabs-karklins (approved):
Looks good

PR #37, opened 2020-10-21
Swap 1129 fetch proposal events from backend

martin-trajanovski:
## Description

This pull request resolves the task: Create an endpoint to fetch all proposal events from the backend.

## Motivation

Previously, the proposal events were hardcoded in the frontend and now we are keeping track of events at only one place(backend) and send them to the frontend so the user officer should be able to select from those when creating the workflow.

## Fixes:

* Task: Create an endpoint to fetch all proposal events from the backend.

## Changes:

Backend adjustments for task resolution


peter-asztalos (approved):
🐱‍🏍

martin-trajanovski:
This should be merged after https://github.com/UserOfficeProject/user-office-backend/pull/33

jekabs-karklins (approved):

mr-field (approved):

PR #45, opened 2020-10-26
Swap 495 as a user i should only see the public status of my proposal

jekabs-karklins:
## Description

User should only be able to see the public status of the proposal;

## Motivation and Context

We do not want to show user in which stage the proposal is in, because that is irrelevant and confusing for the user.
User only needs to see if the proposal is in the draft, submitted, accepted,  rejected or reserved

## How Has This Been Tested

e2e testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-495
As a user I should only see if my proposal is in draft, submitted, rejected, Reserve or accepted
- Remove column submitted,
- In status show DRAFT, SUBMITTED and (ACCEPTED/REJECTED)
- All other statuses between SUBMITTED and ACCEPTED/REJECTED should be simply displayed as SUBMITTED
- Fix that status is always shown as DRAFT

## Changes

Adding new queryable field - public status

New tests:
User should see the right status for the draft proposal
User should see the right status for the submitted proposal


src/queries/ProposalQueries.ts
```diff
@@ -96,4 +103,19 @@ export default class ProposalQueries {
       offset
     );
   }
+
+  async getPublicStatus(agent: UserWithRole | null, id: number) {
```
jekabs-karklins, line 107: @bolmsten This is how the public status is determined

src/datasources/mockups/ProposalDataSource.ts
```diff
@@ -44,43 +45,16 @@ export class ProposalDataSourceMock implements ProposalDataSource {
...
-      1,
-      1,
-      1,
-      1,
-      '',
-      '',
-      false,
-      true
-    );
+    dummyProposal = dummyProposalFactory({ id: 1 });
```
ACLay, line 48: Is this argument needed? The factory uses 1 as the default id

src/datasources/mockups/ProposalDataSource.ts
```diff
@@ -127,27 +101,34 @@ export class ProposalDataSourceMock implements ProposalDataSource {
   }
 
   async get(id: number) {
-    return id === dummyProposal.id ? dummyProposal : null;
+    return (
+      [dummyProposal, dummyProposalSubmitted].find(
```
ACLay, line 105: Should this be `allProposals.find` ?

src/models/Proposal.ts
```diff
@@ -1,9 +1,19 @@
...
   UNSET = 0,
-  ACCEPTED = 1,
-  RESERVED = 2,
-  REJECTED = 3,
+  ACCEPTED,
+  RESERVED,
+  REJECTED,
 }
+
+export enum ProposalPublicStatus {
```
ACLay, line 8: It isn't clear to me what the difference is between `ProposalPublicStatus` and `ProposalEndStatus`. Could some comments be added to clarify what they are intended for?

ACLay (changes requested):

src/datasources/mockups/ProposalDataSource.ts
jekabs-karklins: Yes, from a code perspective not needed, and I thought about passing no arguments, but I believe it's just a little more readable because the other has id:2

src/datasources/mockups/ProposalDataSource.ts
jekabs-karklins: Thanks.

src/models/Proposal.ts
jekabs-karklins: Yes, it should be documented somewhere although at the moment I'm not sure where to place this. 

@martin-trajanovski correct if I'm wrong or you have other thoughts but ProposalEndStatus is something set by the user officer to indicate the final status of the proposal. ProposalPublicStatus is something communicated back to a user. Although they mostly have the same value, they are conceptually different. Also worth noting that PublicStatus is not something stored in the DB but rather a derived value.

martin-trajanovski (approved):

peter-asztalos (approved):

src/models/Proposal.ts
ACLay: Even if it's not the final form of documentation, I still think it would be useful to have some comments around these enums explaining their purposes and what makes them different.

jekabs-karklins: @ACLay I added the descriptions for the 3 proposal statuses we have now, but there is currently work around it, so they are subject to change

ACLay (approved):

PR #48, opened 2020-10-29
SWAP-1143 Artifact factory

peter-asztalos:
## Description

Added support for an external, plug-in PDF creator, on the backend we just collect the data we want in the PDF and we send it to the external service. The current solution waits for the PDF, in the future it should be async and poll based.

## Motivation and Context

The previous solution uses hard coded templates / structure, and it's hard to extend or add support for new PDFs.
The goal here is to collect the information we need send it to an external system which handles the PDF creation.

## How Has This Been Tested

Currently there are no tests, will be added in a different ticket/PR.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1143

## Changes

- new code added which focuses on collecting the data
- feature flag added to use the old system until it's completely removed and tests added

### DIFF
--- a/src/middlewares/pdfFactory.ts
+++ b/src/middlewares/pdfFactory.ts
@@ -69,24 +69,26 @@ const collectSubtemplateData = async (answer: Answer) => {
   const questionaryAnswers: Array<{ fields: Answer[] }> = [];
 
   for (const subQuestionaryId of subQuestionaryIds) {
-    const subquestionarySteps = await questionaryDataSource.getQuestionarySteps(
+    const questionarySteps = await questionaryDataSource.getQuestionarySteps(
       subQuestionaryId
     );
-    if (subquestionarySteps.length === 0) {
-      continue;
-    }
-    const firstStepTopicId = subquestionarySteps[0].topic.id; // NOTE: for now only the first topic
-    const answers = getTopicActiveAnswers(
-      subquestionarySteps!,
-      firstStepTopicId
-    );
 
-    for (const answer of answers) {
-      attachmentIds.push(...getFileAttachmentIds(answer));
-    }
+    const stepAnswers: Answer[] = [];
+
+    questionarySteps.forEach(questionaryStep => {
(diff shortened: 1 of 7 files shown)


src/middlewares/pdfFactory.ts
```diff
@@ -0,0 +1,252 @@
+import contentDisposition from 'content-disposition';
```
peter-asztalos, line 1: Although it's named `pdfFactory` it only generates proposal PDFs, but that should change in the future

martin-trajanovski (approved):

martin-trajanovski:
@peter-asztalos Is this the same as https://github.com/UserOfficeProject/user-office-backend/pull/49? There you just have removed the old pdf generation or I am wrong.

peter-asztalos:
Yes, this comes before https://github.com/UserOfficeProject/user-office-backend/pull/49

peter-asztalos:
@UserOfficeProject/stfc-business-apps

src/middlewares/pdfFactory.ts
```diff
@@ -0,0 +1,252 @@
...
+      subQuestionaryId
+    );
+    if (subquestionarySteps.length === 0) {
+      continue;
+    }
+    const firstStepTopicId = subquestionarySteps[0].topic.id; // NOTE: for now only the first topic
+    const answers = getTopicActiveAnswers(
+      subquestionarySteps!,
+      firstStepTopicId
+    );
```
mr-field, line 82: Why are we getting only the first step? Is there a future task for updating this later on?

peter-asztalos: I copy-pasted that part, @martin-trajanovski @jekabs-karklins may know more about it

jekabs-karklins: Yes, this is fixed now, and all steps should be rendered.

peter-asztalos: should be fixed now

mr-field (approved):

PR #54, opened 2020-11-04
Swap 1132 scheduled events

martin-trajanovski:
## Description

This PR introduces async cron jobs into the system. We have multiple jobs that are going to be executed on a daily basis.

## Motivation and Context

It solves the problem for the need of repetitive jobs that are going to keep proposal workflow running properly.

## How Has This Been Tested

I have created some unit tests for the cron jobs.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1132

## Changes

- Added node-cron package to run async cron jobs that we need. When we start the app it starts all jobs that are going to run on a daily basis. We have 3 jobs for now: checkCallsEndedJob, checkCallsReviewEndedJob, checkCallsSEPReviewEndedJob. There might be a need for more later and this functionality can be easily extended.

- Added some missing dates and flags to the call: startSEPReview, endSEPReview, call_ednded, call_review_ended and call_sep_review_ended.

- Added `sinon` package for better testing of cron jobs using `useFakeTimers` functionality. 


src/asyncJobs/AsyncJobs.spec.ts
```diff
@@ -0,0 +1,151 @@
+/* eslint-disable jest/no-done-callback */
+import 'reflect-metadata';
+import sinon from 'sinon';
```
peter-asztalos, line 3: Why do we use `sinon`? Does `jest` lack some functionality we need?

src/asyncJobs/jobs/checkCallsEnded.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    const updatedCalls = [];
+
+    for (const callThatShouldEnd of callsThatShouldEnd) {
+      const updatedCall = await dataSource.update({
+        ...callThatShouldEnd,
+        callEnded: true,
+      });
+
+      if (!isTestingMode) {
```
peter-asztalos, line 29: I wonder if it would make any sense do this on a higher level and disable event handling in general for some event handlers, like emails, I think the message broker now uses no-op in testing already.

src/asyncJobs/AsyncJobs.spec.ts
martin-trajanovski: I wasn't able to find working functionality for going forward in time using Jest. They have this `advanceTimersByTime` and `useFakeTimers` but it wasn't working for me. Even in the `cron` package that we use for cron jobs I see they are using `sinon` for manipulating the time https://github.com/kelektiv/node-cron/blob/master/tests/cron.test.js. If you can think there is a solution for this using just `Jest` I will be more than happy to implement that and not include another package.

src/asyncJobs/jobs/checkCallsReviewEnded.ts
```diff
@@ -0,0 +1,58 @@
...
+          loggedInUserId: 0,
+        });
+      }
+
+      updatedCalls.push(updatedCall);
+    }
+
+    return updatedCalls;
+  } catch (error) {
+    logger.logError('Checking and ending calls review failed: ', error);
```
peter-asztalos, line 46: Actually, I just remembered, this should be `logException` almost every time when the second parameter is an instance of Error as if you use the logError it doesn't stringify the error properly. I think the logger in general needs some improvements.

src/asyncJobs/jobs/checkCallsEnded.ts
martin-trajanovski: I don't really understand this. You think we can disable the event handling in general on a higher level when in testing mode or?

src/asyncJobs/AsyncJobs.spec.ts
peter-asztalos: Ahh, I see, probably `jest` is not mocking the Node.js timers as good as `sinon`, maybe it's not even their plan, but good to know, I'm fine with `sinon`, at this point it's just a drop in the ocean :D

src/asyncJobs/jobs/checkCallsEnded.ts
peter-asztalos: Yes, when `NODE_ENV` is `test`

peter-asztalos (approved):

src/asyncJobs/AsyncJobs.spec.ts
martin-trajanovski: > Ahh, I see, probably `jest` is not mocking the Node.js timers as good as `sinon`, maybe it's not even their plan, but good to know, I'm fine with `sinon`, at this point it's just a drop in the ocean :D

💧 💦 🌊

src/asyncJobs/jobs/checkCallsEnded.ts
martin-trajanovski: I guess we can do that disable the events in general when testing environment.

jekabs-karklins (approved):

jekabs-karklins:
Just a general note un startAsyncJobs invocation, it could also be done externally with Kubernetes object cronjob (https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/). so the cron is not running on the duo-backend service.

martin-trajanovski:
> Just a general note un startAsyncJobs invocation, it could also be done externally with Kubernetes object cronjob (https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/). so the cron is not running on the duo-backend service.

Yes maybe we can do that. I will try to see how we can do it. Thanks anyway for mentioning it 💯

Scott-James-Hurley (approved):

PR #57, opened 2020-11-05
Swap 1150 refactor Questionary Components

jekabs-karklins:
## Description

This is purely a refactoring task. It does not add nor removes functionality but standardizes how QuestionaryComponents are implemented

Every component in the backend now implements a Question interface and by adding it to the registry it integrates into the questionary system. 

Here is the added documentation for more details.
https://github.com/UserOfficeProject/user-office-backend/wiki/How-to-create-new-questionary-component

## Motivation and Context

Since we expect to add more different types of components and that being the main way to extend functionality, components should be treated as components hiding behind an interface that can be added or removed from the system.

## How Has This Been Tested

- [x] Unit testing
- [x] Manual testing

## Fixes

SWAP-1150

## Changes

Questionary engine
Questionary components


jekabs-karklins:
@UserOfficeProject/stfc-business-apps 
Please look though, there are seemingly a lot of changes but most of it is just moving the same code do a different location :)

src/datasources/mockups/QuestionaryDataSource.ts
```diff
@@ -62,12 +62,15 @@ export const dummyQuestionFactory = (
 export const dummyQuestionTemplateRelationFactory = (
   values?: DeepPartial<QuestionTemplateRelation>
 ): QuestionTemplateRelation => {
-  return new QuestionTemplateRelation(
+  const relation = new QuestionTemplateRelation(
     dummyQuestionFactory(values?.question),
     values?.sortOrder || Math.round(Math.random() * 100),
     values?.topicId || Math.round(Math.random() * 10),
-    new BooleanConfig()
+    (values?.config as any) || new BooleanConfig(),
```
martin-trajanovski, line 69: Can't we define those `any` types somehow or it is a bit hard to do it?

martin-trajanovski (approved):

src/datasources/mockups/QuestionaryDataSource.ts
jekabs-karklins: It's because dummyQuestionTemplateRelationFactory accepts an initial value as of type DeepPartial<QuestionTemplateRelation>, meaning that any property including nested property is optional. The idea for dummyQuestionTemplateRelationFactory() is that you only specify the values you want and the remaining will be automatically filled with a dummy value, and that is very useful. Makes code shorter.
The problem is that for the config it's a union datatype, so I can't really know what data type is passed.

But with a little bit of TypeScript magic I made so that config is either specified completely or omitted at all.

jekabs-karklins:
@RasmiaKulan can you please review this? Thanks.

src/models/ProposalModelFunctions.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+    config: createConfig<BooleanConfig>(DataType.TEXT_INPUT, {
+      required: false,
+    }),
+  });
+  expect(isMatchingConstraints(question, 'text')).toBe(true);
+  expect(isMatchingConstraints(question, '')).toBe(true);
+});
+
+it('Dependencies should be sattisfied if value matches', async () => {
```
RasmiaKulan, line 47: There is a typo in spelling out 'satisfied'

jekabs-karklins: ✓

jekabs-karklins:
@RasmiaKulan I'm also waiting for "approve" before I can merge this into the develop

RasmiaKulan (approved):

PR #60, opened 2020-11-10
Swap 1166 question order not saved

martin-trajanovski:
## Description

This is a bugfix and improvement. This changes the way we are ordering and sorting the questions inside topics and topics inside a templates.

## Motivation and Context

This is bugfix and improvement in the functionality for sorting and soring sort values inside database. Previously we had problems with sorting questions and topics because if new question was added in between two questions we were just inserting the new question and not updating the sort_order for questions after it so two or more questions were ending up on the same position which wasn't good for the positioning on the frontend. Now that is solved with updating the other items in the step when you add/reorder a question or a topic.

## How Has This Been Tested

It has unit and e2e tests.

## Fixes

Fixes https://jira.esss.lu.se/browse/SWAP-1166

## Changes

When question is reordered or added to a topic now we update all other items in that topic to sort_order them correctly. I am ordering them firstly and then I do upsert in to the database to update(existing)/insert(new) items.


peter-asztalos (approved):

jekabs-karklins (approved):

ACLay (changes requested):
At a high level, I'm unclear why was it chosen to switch to floats instead of updating the index values on subsequent questions/sections, and I think doing so introduces a problem.

With floats, reordering by using inbetween values could lead to the order values coalescing, and clashing once they get too close. With 8 decimals of precision I reckon this could happen after log2(10^8) ≈ 27 reorderings.

I could see this happening down the line if a template gets repeatedly copied and adapted between rounds. I don't think this sort of error could occur if we use integers and update the sort indexes of later items in the list.

martin-trajanovski:
> At a high level, I'm unclear why was it chosen to switch to floats instead of updating the index values on subsequent questions/sections, and I think doing so introduces a problem.
> 
> With floats, reordering by using inbetween values could lead to the order values coalescing, and clashing once they get too close. With 8 decimals of precision I reckon this could happen after log2(10^8) ≈ 27 reorderings.
> 
> I could see this happening down the line if a template gets repeatedly copied and adapted between rounds. I don't think this sort of error could occur if we use integers and update the sort indexes of later items in the list.

martin-trajanovski:
> At a high level, I'm unclear why was it chosen to switch to floats instead of updating the index values on subsequent questions/sections, and I think doing so introduces a problem.
> 
> With floats, reordering by using inbetween values could lead to the order values coalescing, and clashing once they get too close. With 8 decimals of precision I reckon this could happen after log2(10^8) ≈ 27 reorderings.
> 
> I could see this happening down the line if a template gets repeatedly copied and adapted between rounds. I don't think this sort of error could occur if we use integers and update the sort indexes of later items in the list.

The main issue was that we were not updating the sort indexes of later items in the list at all. I tried to research a bit and I thought that maybe this will be the most efficient solution because we wont need to update anything just insert the reordered or new item at the right position. I have seen some examples where in Trello boards they use this kind of approach. I was thinking that it will be good for our needs as well but now I see your point and I guess that it might be better to use integers and just update the entire list of items after the one changed. I will rethink and discuss with others and we will come with conclusion what works the best for us. Thank you for the comment anyway.

src/mutations/TemplateMutations.ts
```diff
@@ -281,15 +318,54 @@ export default class TemplateMutations {
...
     args: UpdateQuestionTemplateRelationArgs
-  ): Promise<Template | Rejection> {
+  ): Promise<Template | Rejection | null> {
+    const dataToUpsert = await this.getQuestionsDataToUpsert(args);
+
     return this.dataSource
-      .updateQuestionTemplateRelation(args)
-      .then(steps => steps)
+      .upsertQuestionTemplateRelations(dataToUpsert)
+      .then(data => data)
```
ACLay, line 368: What are these `.then(data => data)` lines for? They don't look like they do anything.

src/populate/index.ts
```diff
@@ -144,7 +148,11 @@ const createTemplates = async () => {
   for (const template of templates) {
     await execute(() => {
       return templateDataSource.createTopic({
-        sortOrder: 0,
+        sortOrder: faker.random.number({
```
ACLay, line 151: I'm guessing that this file is to create some dummy data in test/development environments?
Either way, it looks like this sortOrder generator wasn't updated to generate integers alongside the other one on line 184

ACLay (changes requested):

src/mutations/TemplateMutations.ts
martin-trajanovski: Good catch maybe we save some lines if we remove them

src/populate/index.ts
martin-trajanovski: I will check and update this as well

ACLay (approved):

PR #65, opened 2020-11-12
bump validation version

martin-trajanovski:
## Description

This is just to update the validation version package and the changes for resolving the issue are there.


peter-asztalos (approved):

src/datasources/mockups/UserDataSource.ts
```diff
@@ -127,8 +127,8 @@ export const dummyUserNotOnProposal = new User(
   'Facilitator',
   'Tyrique41@hotmail.com',
   true,
-  '1-272-760-1466 x03877',
-  '174-603-1024',
+  '+46700568256',
```
jekabs-karklins, line 130: Why are these changes needed? Did we change the acceptable phone format?

martin-trajanovski: Yes I changed the phone input validation some time ago but it wasn't working well and now I fixed that a bit. The previous number wasn't valid one and the tests were failing all the time. We should have a tests as well trying to set the invalid number and see if validation works properly. I can write some on the frontend to improve user create/update tests.

PR #59, opened 2020-11-09
SWAP-1201 CI/CD Improvements

peter-asztalos:
## Description

This PR updates the Github action and Docker configurations to improve the overall speed, size of the images and makes action files consistent.

## Motivation and Context

The previous solution didn't use any caching.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1201

## Changes

- save intermediate layers after each build and use them if there were no changes
- docker ignore updated to provide a better caching experience 
- use multi-stage builds to reduce image size (~290->~160mb)
- package.json updated to use only proper dependencies in each environment
- e2e test related configurations

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/52

## Tests included/Docs Updated?

- [ ] I have added tests to cover my changes.
- [ ] All relevant doc has been updated

PR #76, opened 2020-11-17
Improve Windows support

mr-field:
Ref UserOfficeProject/stfc-user-office-project#49

## Description
The first commit adds a very basic Dockerfile that can be used while developing the app. Instead of performing a clean build, it just expects the code to be mounted into the image and starts the app in debug mode with automatic recompilation. As a consequence, it's not necessary to rebuild the image with every code change - it can just be started once and the code will be recompiled as necessary (note that this requires the `-i` flag so that Docker keeps STDIN open).

I've also added a new `test` command to set the environment variables using PowerShell, as Windows doesn't natively understand the `env` command and we'll need to test our changes locally.

The second commit applies the changes in #59 to the Windows "prod" Dockerfile.

## Motivation and Context
This will make develepment much faster for Windows developers who prefer to run the app in Docker

## How Has This Been Tested
The backend tests have been run in the new Docker container by running `docker exec npm.cmd run test` while it was up.

## Changes
The workflow for developing on Windows using Docker will be slightly different after this (the service just needs to be started once).

I will document the new workflow on the STFC wiki, as it uses our Compose files which aren't (yet?) public.


mr-field:
Closing as the branch name contains a `#`, which isn't a valid character for a Docker tag

PR #79, opened 2020-11-18
Swap 1199 filtering based on statuses

martin-trajanovski:
## Description

Filtering the proposals by the statuses from the workflow.

## Motivation and Context

This is the solution for better following of the proposal workflow. If you as a Instrument Scientist need to see only proposals that are for feasibility review they are filtered by default on that.

## How Has This Been Tested

With unit and e2e tests on the frontend.

## Fixes

It fixes https://jira.esss.lu.se/browse/SWAP-1199

## Changes

More on the changes are on the backend and here it is only adding and filtering by `proposalStatusId`


peter-asztalos (approved):

src/queries/ProposalSettingsQueries.ts
```diff
@@ -15,7 +15,7 @@ export default class ProposalSettingsQueries {
     return proposalStatus;
   }
 
-  @Authorized([Roles.USER_OFFICER])
+  @Authorized()
```
jekabs-karklins, line 18: Is this intentional?

jekabs-karklins (approved):
Looks good

src/queries/ProposalSettingsQueries.ts
martin-trajanovski: Yes it is because now when I try to filter based on proposal statuses I have a filter dropdown where I want to show all statuses that proposal can be filtered by. It can be called by instrument scientist as well or SEP chair.

EdwardHaynes (approved):

PR #81, opened 2020-11-18
Adding liveness and readiness checks

jekabs-karklins:
## Description

Adding two new endpoints for kubernetes integration
- readiness
- liveness

## Motivation and Context

With these endpoints it will help Kubernetes to decide on then, service is ready or has crashed and needs to be restarted 

## How Has This Been Tested

Manual testing

## Fixes

N/A

## Changes

Adding new data source, new query, and two new paths

## Depends on

N/A


src/datasources/postgres/SystemDataSource.ts
```diff
@@ -0,0 +1,25 @@
+import { logger } from '@esss-swap/duo-logger';
+
+import { SystemDataSource } from '../SystemDataSource';
+import database from './database';
+
+export default class PostgreSystemDataSource implements SystemDataSource {
+  async connectivityCheck(): Promise<boolean> {
+    try {
+      return database
+        .raw('select 1+1 as result')
```
peter-asztalos, line 10: I think this could be re-used for both, as having a working database connection is essential to call the app healthy, but I think we need only one endpoint for it, as they both want to know the same thing (in our use case) but with different timing, like for readiness you may want to initially wait longer before checking the status, while if the app is already running you just want to periodically check its state.

src/middlewares/healthCheck.ts
```diff
@@ -0,0 +1,11 @@
+import express, { Request, Response } from 'express';
+
+const router = express.Router();
+
+router.get('/healthz', (req: Request, res: Response) => {
```
peter-asztalos, line 5: I used `/health-check` for other components (except the gateway which uses the Apollo's endpoint), I think we should be consistent here and decide on the endpoint:
- /healthz
- /health-check
- /.well-known/apollo/server-health (or something similar)
- something else

src/datasources/postgres/SystemDataSource.ts
jekabs-karklins: I think we should not have a liveness probe check the database, because if the database goes down we don't want the duo-backend to be terminated and recreated. It should only fail the readiness probe.

jekabs-karklins: But we can talk about it

src/middlewares/healthCheck.ts
jekabs-karklins: Yes, I went with K8 recommendations, we should discuss it too.

TickleThePanda (approved):
Apart from the discussion about the best practice for endpoint names, this looks good to me.

Am I right in understanding that `<endpoint>z` is to stop name collisions from other endpoints? If that's the case, I don't mind what we call the endpoints (as long as there's consistency) as I don't think there's a risk of collision here.

peter-asztalos:
They use `/healthz` in their [example](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request), maybe the additional `z` has something to do with collisions as you mentioned, or it's a foo bar thing. I think that's why the Apollo lib uses the `/.well-known/...` ? 

But yeah, the main goal is to be consistent.

jekabs-karklins:
> They use `/healthz` in their [example](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request), maybe the additional `z` has something to do with collisions as you mentioned, or it's a foo bar thing. I think that's why the Apollo lib uses the `/.well-known/...` ?
> 
> But yeah, the main goal is to be consistent.

They are called z-pages and postfixed with z to avoid conflicts.

PR #84, opened 2020-11-20
SWAP-1183 Excel export

peter-asztalos:
## Description

This PR adds support for generating XLSX on the server-side based on the input coming from the User office.
It adds support for exporting Proposals and SEPs in XLSX.

## Motivation and Context

Remove the xlsx lib from the FE (saves roughly 1.2MB is bundle size) and generate the files on the server-side.

## How Has This Been Tested

- [x] unit test in Factory repo
- [x] e2e test in the FE repo

## Fixes

https://jira.esss.lu.se/browse/SWAP-1183

## Changes

- existing proposal export logic moved
- new export logic for SEP added
- little refactoring/restructuring

## Depends on

https://github.com/UserOfficeProject/user-office-factory/pull/11
https://github.com/UserOfficeProject/user-office-frontend/pull/99


martin-trajanovski (approved):

src/factory/service.ts
```diff
@@ -0,0 +1,77 @@
...
+export enum PDFType {
+  PROPOSAL = 'proposal',
+  SAMPLE = 'sample',
+}
+
+export type MetaBase = { collectionFilename: string; singleFilename: string };
+export type XLSXMetaBase = MetaBase & { columns: string[] };
+
+// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+const ENDPOINT = process.env.USER_OFFICE_FACTORY_ENDPOINT!;
```
jekabs-karklins, line 27: I think we should either thrown an error or handle the case when the endpoint is not provided by returning an error to the client then the request cant be handled because of the missing endpoint

peter-asztalos: Well, yeah, if the endpoint doesn't exist then the logs will contain the error, but it's impossible to inform the FE what was the error unless we create an endpoint which is used for checking if the endpoint is defined. 

This should be defined every time, even if it doesn't point anywhere, maybe it would be better to just crash the backend if it doesn't exist.

jekabs-karklins: Maybe we should go with "Fail Fast, Fail Often"

jekabs-karklins (approved):

src/factory/xlsx/proposal.ts
```diff
@@ -0,0 +1,86 @@
...
+  );
+
+  const proposer = await baseContext.queries.user.get(
+    user,
+    proposal.proposerId
+  );
+
+  if (!proposer) {
+    throw new Error(
+      `Proposal with ID '${proposalId}' not found, or the user has insufficient rights`
```
mr-field, line 55: I think this should say something like "Proposer not found", as the proposal should be OK if we made it this far

src/factory/xlsx/sep.ts
```diff
@@ -0,0 +1,201 @@
...
+        callId,
+        instrumentId: instrument.id,
+      });
+    })
+  );
+
+  const instrumentsProposals = await Promise.all(
+    instrumentsSepsProposalIds.map(sepProposalIds => {
+      if (!sepProposalIds) {
+        throw new Error('todo');
```
mr-field, line 98: Is "todo" a placeholder as the functionality hasn't been implemented yet? If not, it would be useful to make it a full error message

src/middlewares/factory/xlsx.ts
```diff
@@ -0,0 +1,106 @@
...
+      { data, meta },
+      res,
+      next
+    );
+  } catch (e) {
+    next(e);
+  }
+});
+
+router.get(`/${XLSXType.SEP}/:sep_id/call/:call_id`, async (req, res, next) => {
```
mr-field, line 65: I've noticed that the `/${XLSXType.PROPOSAL}` endpoint stops if the user is not a user officer, but this endpoint doesn't. Is this expected?

mr-field (commented):
Look good, just a couple of comments

src/middlewares/factory/xlsx.ts
peter-asztalos: Yes, SEP members should be also able to access it, so the permission check is done inside the `collectSEPlXLSXData` by the queries.

PR #86, opened 2020-11-24
Swap 1235 default proposal statuses

martin-trajanovski:
## Description

Adding more default proposal statuses and added `short_code` field to be able to use name in a more human readable way. 

## Motivation and Context

This is solving the need for more default proposal statuses that are going to be used inside proposal workflows and need for more human readable name everywhere we use the proposal statuses.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1235

## Changes

Changes in the database to add new field and some refactoring to be able to use it.

## Depends on

https://github.com/UserOfficeProject/user-office-lib/pull/17


src/mutations/ProposalSettingsMutations.ts
```diff
@@ -59,22 +56,14 @@ export default class ProposalSettingsMutations {
     agent: UserWithRole | null,
     args: UpdateProposalStatusInput
   ): Promise<ProposalStatus | Rejection> {
-    // TODO: Find better way of preventing update and remove on default proposal statuses.
-    if (args.id < 10) {
-      return rejection('NOT_ALLOWED');
-    }
```
peter-asztalos, line 65: The `updateProposalStatus` doesn't include the `is_default` check, does that mean that we allow update but not delete, or you forgot to add the check?

martin-trajanovski: We allow partial update yes. Only on `name` and `description` but not on the `short_code`

peter-asztalos (approved):

jekabs-karklins (approved):

PR #89, opened 2020-11-25
SWAP-1228 Table is missing data after SEP deactivation

peter-asztalos:
## Description

This PR aims to fix an issue related to the Proposals table, where the connected SEP doesn't show up.

## Motivation and Context

The Proposals table uses a view to display the relevant information we need, this view used an obsolete field to populate the SEP code.
This PR removes the unused field and updates the view.

## How Has This Been Tested

- [x] Manual testing
- [x] e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1228

## Changes

- `proposal_view_table` updated
- obsolete `sep_id` field removed
- field names updated

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/107
https://github.com/UserOfficeProject/user-office-backend/pull/86

### DIFF
--- a/db_patches/0073_UpdateProposalView.sql
+++ b/db_patches/0073_UpdateProposalView.sql
@@ -1,0 +1,47 @@
+DO
+$$
+BEGIN
+    IF register_patch('UpdateProposalView.sql', 'Peter Asztalos', 'Use SEP_Proposals for joining SEP', '2020-11-25') THEN
+
+        DROP VIEW proposal_table_view;
+
+        ALTER TABLE proposals DROP COLUMN sep_id;
+
+        CREATE VIEW proposal_table_view
+        AS
+        SELECT  p.proposal_id AS id,
+                p.title,
+                p.status_id AS proposal_status_id,
+                ps.name AS proposal_status_name,
+                ps.description AS proposal_status_description,
+                p.short_code,
+                p.rank_order,
+                p.final_status,
+                p.notified,
+                t.time_allocation,
+                t.status AS technical_review_status,
+                i.name AS instrument_name,
+                c.call_short_code,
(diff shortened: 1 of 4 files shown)


db_patches/0073_UpdateProposalView.sql
```diff
@@ -0,0 +1,47 @@
+DO
+$$
+BEGIN
+    IF register_patch('UpdateProposalView.sql', 'Peter Asztalos', 'Use SEP_Proposals for joining SEP', '2020-11-25') THEN
+
+        DROP VIEW proposal_table_view;
+
+        ALTER TABLE proposals DROP COLUMN sep_id;
```
peter-asztalos, line 8: We use `SEP_Proposals` for connecting the proposal to SEP.
I couldn't find any reference to it except the `proposal_table_view` view.

NOTE: should we try to create the missing `SEP_Proposal` based on this field?

martin-trajanovski: You think about creating SEP_Proposal based on this `sep_id` on `proposals`? I think that for every proposal that was added to SEP the SEP_Proposal record is created in the database but we were not updating the `sep_id` on proposal itself. This is just a redundant data and can be removed I guess as you have done and we just keep the `SEP_Proposals` as the only source where we get proposals assigned to SEP.

martin-trajanovski (approved):

db_patches/0073_UpdateProposalView.sql
peter-asztalos: I see, I don't know the history of `sep_id` if it was used by any chance in critical parts, if it wasn't then it makes things easier as we can just simply drop it.

PR #96, opened 2020-11-30
SWAP-1157 #resolved

jekabs-karklins:
## Description

Store file id as UUID instead of bigint

## Motivation and Context

File ID used to be BIGINT (2^53 - 1), which has conversion problems to JS number which max number is smaller than Postgres BIGINT. This takes care of storing file id as UUID which is converted to string in JS.

Also, the file ID is a long UUID which is hard to guess as we don't have authentication for downloading files for now. 

## How Has This Been Tested

- [x] Manual testing

## Fixes

SWAP-1157

## Changes

Altering database
Migrating data

## Depends on

N/A

### DIFF
--- a/db_patches/0074_MigrateFiesToUUID.sql
+++ b/db_patches/0074_MigrateFiesToUUID.sql
@@ -1,0 +1,86 @@
+DO
+$$
+BEGIN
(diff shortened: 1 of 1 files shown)


martin-trajanovski (approved):
A bit more complicated migration but it looks good to me.

peter-asztalos (approved):

PR #108, opened 2020-12-15
Add external auth support

mr-field:
Part of UserOfficeProject/stfc-user-office-project#44

## Description
See initial design at UserOfficeProject/stfc-user-office-project#48

This PR adds a new mutation which can be used to login a user with their STFC credentials. It follows the same logic explained in the inital design linked above - the new mutation takes an STFC auth token, uses it to get the user details, and returns a new JWT encoding the user details and roles.

Note that at this stage none of the user details are copied from the STFC user to the proposals user, only the user number is used to create a dummy record in the proposals DB.

The initial design had a note about making the implmentation more general by introducing something like an `ExternalAuthDataSource` interface, then moving the STFC-specific code to an implementation of this general inerface. While I'd still like to do that, I think we should wait until we have ouw own implementation of the `UserDataSource` to see more clearly how the two will interact.

## Motivation and Context
This will allow STFC users to sign into the proposals app with their existing accounts.

## How Has This Been Tested
I've added 2 Jest tests for the new mutations. The change has also been tested manually by giving different Session IDs to the new frontend endpoint.

## Changes
Other than the new mutation and related code, the only change in this PR is a small alteration to some of the "get" methods in the Postgres `UserDataSource`. They now return `null` when a user isn't found, rather than crashing when trying to convert the undefined raw result to a model object. This seems to fit their type signature better, as they are meant to return `Promise<User|null>`.

## Depends on
https://github.com/UserOfficeProject/user-office-frontend/pull/139


package.json
```diff
@@ -86,6 +88,7 @@
     "@types/pg-large-object": "^2.0.2",
     "@types/react-beautiful-dnd": "^11.0.5",
     "@types/request-promise": "^4.1.46",
+    "@types/sax": "^1.2.1",
```
ACLay, line 91: I'm not too familiar with node dependencies. Why is this needed? I can't see any code changes here that import anything from sax.

mr-field: I'm not too sure myself. Without the import, running `npm run lint` results in this error:
```
node_modules/soap/lib/wsdl/index.d.ts:1:22 - error TS7016: Could not find a declaration file for module 'sax'. 'C:/programming/user-office-backend/node_modules/sax/lib/sax.js' implicitly has an 'any' type.
  Try `npm install @types/sax` if it exists or add a new declaration (.d.ts) file containing `declare module 'sax';`
```

The `soap` package is required by `@stfc-user-programme/uows_client_generator`, but it's not used directly in this app. The only explanation I can think of is that the `soap` [package.json](https://github.com/vpulim/node-soap/blob/master/package.json) requires `sax` at runtime, but `@types/sax` only at build time. This may mean that Typescript projects which import `soap` have to provide their own dev import for `@types/sax`, since it's not required by the package itself. I could be wrong though, I'd appreciate suggestions by people who know NPM better.

peter-asztalos: This is something that we could improve on the `@stfc-user-programme` side by adding linter, but before that, I think this should be added to the `@stfc-user-programme/uows_client_generator` package.json which actually uses the `soap` lib.

peter-asztalos (approved):

ACLay (approved):

martin-trajanovski (approved):

PR #123, opened 2021-01-11
Swap 1253 submit instrument by SEP

martin-trajanovski:
## Description

Now every proposal on the instrument has a flag submitted and when we submit instrument from the SEPs we are submitting it for each and every proposal that is in the actual SEP and is attached to that instrument. Previously we were submitting the instrument by call and if we had proposals under different SEPs and the same instrument it was getting submitted for all SEPs which was incorrect.

## Motivation and Context

We have better control when instrument is submitted and it fixes the incorrect way of doing it for all SEPs.

## How Has This Been Tested

E2e and unit tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1253

## Changes

Adding new submitted flag inside instrument_has_proposals table and using the right logic to change it.


src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -268,6 +269,34 @@ export default class PostgresInstrumentDataSource
...
+      );
+
+      const allProposalsOnInstrumentSubmitted = allProposalsOnInstrument.reduce(
+        (sum, next) => sum && (next.instrumentSubmitted as boolean),
+        true
+      );
+
+      instrumentsWithSubmittedFlag.push({
+        ...instrument,
+        submitted: allProposalsOnInstrumentSubmitted || false,
```
peter-asztalos, line 293: I think the default `false` is not required here

peter-asztalos (approved):

src/datasources/postgres/InstrumentDataSource.ts
jekabs-karklins: Maybe 
`const allProposalsOnInstrumentSubmitted = allProposalsOnInstrument.every(item => item.instrumentSubmitted)`
 would be more readable in this case

jekabs-karklins (approved):

PR #125, opened 2021-01-14
SWAP-756 adding feature toggle

jekabs-karklins:
## Description

Adding functionality to turn on/off features

## Motivation and Context

This feature has multiple applications. It allows releasing the frontend with the feature disabled and then enabling it afterward.
This feature also can be used to differentiate between features that are enabled across facilities.

## How Has This Been Tested

- [x] added unit tests 

## Fixes

SWAP-756 add feature toggle

## Changes

New table "features", and new query

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/183


peter-asztalos (approved):
Looks good, but I think the shipments e2e tests are failing because they are off by default, we might need an endpoint for e2e testing where we can turn features on/off (like resetDB which runs before specific tests starts).

jekabs-karklins:
> Looks good, but I think the shipments e2e tests are failing because they are off by default, we might need an endpoint for e2e testing where we can turn features on/off (like resetDB which runs before specific test starts).

OK, I have enabled the feature in the DB by adding a new entry

martin-trajanovski (approved):

Scott-James-Hurley (approved):

PR #129, opened 2021-01-18
SWAP-1182 SEP time allocation

peter-asztalos:
## Description

This PR has two big parts:

1. added support for SEP time allocation, which can "override" the technical review time allocation
2. fixed (mostly role-related) bugs in the SEP module (see changes for a detailed list)

## Motivation and Context

SEP Chair and Secretary currently have no way to "override" the allocated time, which is important e.g. for calculating available time.

There are many inconsistencies or role-related bugs in the SEP module which also lack e2e tests.

## How Has This Been Tested

- [x] e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1182

## Changes

- added support to "override" technical review time allocation,
- disallow SEP Reviewers to access detailed proposal view in 'Meeting components' if they are not assigned to the proposal,
- disallow SEP Chair/Secretary to modify Proposal in 'Meeting components' if the instrument is already submitted,
- allow SEP Chair/Secretary to modify Proposal in 'Meeting components' if the instrument is not submitted yet,
- disallow SEP Reviewers to access other SEP Reviewers' review,
- allow SEP Chair/Secretary to access/modify reviews,
- allow SEP Chair/Secretary to (un)assign SEP Reviewer to proposals,
- disallow SEP Chair/Secretary to (un)assign SEP Chair/Secretary members,
- some component state updated after unmount errors were fixed

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/189


src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -495,4 +495,18 @@ export default class PostgresInstrumentDataSource
         return result?.count === '1';
       });
   }
+
+  async isProposalInstrumentSubmitted(proposalId: number): Promise<boolean> {
+    return database('instrument_has_proposals')
+      .select()
+      .where('proposal_id', proposalId)
```
martin-trajanovski, line 502: I think we should select by `proposalId` and `instrumentId` as well. Because I guess if one proposal is assigned on two instruments than we get the first record and maybe we would like to check second one where `instrumentId` is different.

martin-trajanovski (approved):

src/datasources/postgres/InstrumentDataSource.ts
peter-asztalos: I thought the same, but currently, you can assign only one instrument to a proposal.

martin-trajanovski: Yes then it is fine I guess. I wasn't sure if we already have that limitation of assigning proposals to instrument.

jekabs-karklins (approved):

Scott-James-Hurley (approved):

PR #130, opened 2021-01-19
Swap 1295 question multiple dependencies

martin-trajanovski:
## Description

Now we are able to add multiple dependencies on a single question. They are all compared using AND operator currently.

## Motivation and Context

There is a need and use cases where it is easier to create template if we have one question depending on multiple other questions.

## How Has This Been Tested

Unit tests and e2e tests on the frontend.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1295

## Changes

Moving question dependencies in their own database table, data migration and support for all the logic to get everything to work together.


jekabs-karklins (approved):
Looks very good

Scott-James-Hurley (approved):

PR #132, opened 2021-01-20
Stfc user office project#4 send smtp mail

Scott-James-Hurley:
ref: UserOfficeProject/stfc-user-office-project#4

This adds the ability to send emails using the SMTP protocol, though Sparkpost is still the default. This is so STFC users of ESS can be sent emails. The host, port, sending address and username and password for authentication are passed via environment variables. Sparkpost has been wrapped in a class to make it a bit easier to use. The SMTP mail functionality has been testing using Jest.


src/eventHandlers/emails/submit/html.pug
```diff
@@ -0,0 +1,15 @@
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8">You have been included on a proposal for experimental time at the ESS by <b>#{piPreferredname} #{piLastname}</b><p></p><p>As the Principal Investigator on an ESS proposal you have specific responsibilities for the experiment.
```
TickleThePanda, line 1: While we've been informally calling this software the "ESS proposal software", for us, we're intending it to be used at our facilities (ISIS and the different areas of the CLF). For now, could you remove mentions of the ESS in here?

I think it would also be useful to replace `<br>` with `<p>` (you only need 1 `<p>` for each paragraph rather than the multiple `<br>`s) and `<b>` with `<strong>` as that's best practice for "semantic HTML".

TickleThePanda (changes requested):

src/eventHandlers/emails/submit/html.pug
```diff
@@ -1,15 +1,20 @@
-<meta http-equiv="Content-Type" content="text/html; charset=utf-8">You have been included on a proposal for experimental time at the ESS by <b>#{piPreferredname} #{piLastname}</b><p></p><p>As the Principal Investigator on an ESS proposal you have specific responsibilities for the experiment. 
-<br>
-You can read about these responsibilities here: ESS_TERMS_OF_ACCESS_URL.
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8"><p>You have been included on a proposal for experimental time at ISIS/CLF by <strong>#{piPreferredname} #{piLastname}</strong></p>
+
+<p>As the Principal Investigator on an ESS proposal you have specific responsibilities for the experiment.<br>
```
TickleThePanda, line 3: I don't think we should have any `<br>` tags at all in here. For now, if that means there's poor style, I don't think it's an issue.

TickleThePanda (changes requested):

src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -0,0 +1,94 @@
...
+      id: Math.random()
+        .toString(36)
+        .substring(7),
+    };
+
+    if (process.env.NODE_ENV !== 'production') {
+      sendMailResults.id = 'test';
+    }
+
+    options.content.template_id =
```
TickleThePanda, line 56: The directory for where the templates come from should be externalised. Probably in an env variable?

TickleThePanda (changes requested):

src/eventHandlers/MailService/SMTPMailService.ts
Scott-James-Hurley: Changed SMTPMailService.ts line 57 to use an environment variable.

mr-field:
@Scott-James-Hurley Could yo try to solve the merge conflicts and push this branch again to see if that unlocks the tests?

Scott-James-Hurley:
@mr-field I've merged it with the latest version of develop. The tests pass but the build job fails.

src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -0,0 +1,92 @@
...
+    const sendMailResults: SendMailResults = {
+      total_rejected_recipients: 0,
+      total_accepted_recipients: 0,
+      id: Math.random()
+        .toString(36)
+        .substring(7),
+    };
+
+    if (process.env.NODE_ENV !== 'production') {
+      sendMailResults.id = 'test';
```
mr-field, line 51: Since this case is specifically for unit testing, could the check be changed to `process.env.NODE_ENV === 'test'`? That way it won't affect the dev environment once we have it

mr-field (changes requested):
The build is failing because the build script tries to use the branch name as a tag for the built Docker image. Unfortunately, this branch has a `#` in the name which isn't a valid character for Docker tags. The only way around it would be to close this PR, rename the branch, and create a new PR from the renamed branch. However, I think we should finish the review first to avoid confusion

src/eventHandlers/email.ts
```diff
@@ -1,22 +1,36 @@
...
+import { SMTPMailService } from './MailService/SMTPMailService';
+import { SparkPostMailService } from './MailService/SparkPostMailService';
 
 const options = {
   endpoint: 'https://api.eu.sparkpost.com:443',
 };
-const client = new SparkPost(process.env.SPARKPOST_TOKEN, options);
+
+if (process.env.NODE_ENV !== 'production') {
+  dotenv.config();
```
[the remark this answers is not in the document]
Scott-James-Hurley: yeah you right

package.json
```diff
@@ -50,9 +50,10 @@
     "bcryptjs": "^2.4.3",
     "bluebird": "^3.7.2",
     "class-validator": "^0.13.1",
+    "cache-pug-templates": "^2.0.3",
     "cookie-parser": "^1.4.5",
     "cron": "^1.8.2",
-    "dotenv": "^8.2.0",
```
mr-field, line 55: I think this is still referenced in `index.ts`. Not sure if it's actually used, but removing it will likely make "clean" builds fail

mr-field (changes requested):
Thanks, one last thing

mr-field (approved):

mr-field:
@TickleThePanda Could you review this again before we reopen the PR with a renamed branch?

TickleThePanda (approved):

PR #135, opened 2021-01-22
SWAP-1294 #resolved

martin-trajanovski:
## Description

Adding new filter parameter `shortCode` when querying the proposals.

## Motivation and Context

There is a need to be able to get the proposal by `shortCode`.

## How Has This Been Tested

Manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-1294

## Changes

Just adding support to filter proposals by `shortCode`

### DIFF
--- a/src/datasources/postgres/ProposalDataSource.ts
+++ b/src/datasources/postgres/ProposalDataSource.ts
@@ -188,11 +188,13 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
           );
         }
 
-        if (filter?.shortCode) {
-          query.where(
-            'proposal_table_view.short_code',
-            'like',
-            `%${filter?.shortCode}%`
+        if (filter?.shortCodes) {
+          const filteredAndPreparedShortCodes = filter?.shortCodes
+            .filter(shortCode => shortCode)
+            .join('|');
+
+          query.whereRaw(
+            `proposal_table_view.short_code similar to '%(${filteredAndPreparedShortCodes})%'`
           );
         }
       })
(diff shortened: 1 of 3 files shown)


src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -187,6 +187,14 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
             filter?.proposalStatusId
           );
         }
+
+        if (filter?.shortCode) {
+          query.where(
+            'proposal_table_view.short_code',
+            'like',
```
peter-asztalos, line 194: Is `short_code` transformed either to upper or lower case (uses the same case everywhere)?
If not I might consider using `ilike` to make the search case-insensitive.

peter-asztalos (approved):

src/datasources/postgres/ProposalDataSource.ts
martin-trajanovski: I think it is only numbers so that is why I went with just `like`

jekabs-karklins: As we talked in the morning it might be also a good idea to have a parameter short_codes with input a type of array, in case they want to do a bulk search.

jekabs-karklins (approved):

src/datasources/postgres/ProposalDataSource.ts
martin-trajanovski: I will try to implement support for that as well.

PR #136, opened 2021-01-22
Authorise external STFC users

simonfernandes:
## Description

This allows a user to be authorised via STFC's User Office Web Service (UOWS). This has been done by implementing a UserDataSource specific to STFC that fetches roles from the UOWS. The complete `StfcUserDataSource` has been implemented to fetch any other needed data from the UOWS, too.

Some functions (e.g. `addUserRole()`) are set to throw since we only want to support it through our systems. In these cases we would want to disable/hide these options to the user, but this will be done in a separate issue as it's lower priority and will require more thought and frontend knowledge.

## Motivation and Context

Currently the app authenticates an STFC user but grants the user all roles, and fetches everything from the database instead of going off to the UOWS.

## How Has This Been Tested

All unit tests pass, and I have added a couple that test the new data source.

All pages seem to work while switching between User and User Officer role. Users can be viewed, proposals can be saved and submitted.

I have also tested it still works when none of the STFC-related environment variables are set (using the standard login method with Postgres user data source).

## Fixes

Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/45
Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/72

## Changes

Adds an STFC user data source, which is chosen for authentication and general use by the presence of an environment variable. When authenticating an external user, roles are now fetched using the data source.

## Depends on

Depends on a docker-orchestration PR (internal STFC repo) that adds the required environment variables: https://github.com/isisbusapps/docker-orchestration/pull/31

- [x] I have added tests to cover my changes (some).


package.json
```diff
@@ -23,8 +23,7 @@
     "start": "tsc && node ./build/index.js",
     "apitest": "newman run api_tests/tests.json --environment api_tests/environment.json",
     "debug": "rm -R -f ./build && tsc && DATABASE_URL=postgres://duouser:duopassword@127.0.0.1:5432/duo secret=qMyLZALzs229ybdQXNyzYRdju7X784TH tokenLife=7d baseURL=localhost:3000 NODE_ENV=development SPARKPOST_TOKEN=insertokenhere GRAYLOG_SERVER=127.0.0.1 GRAYLOG_PORT=12202 ts-node-dev --inspect --respawn -- ./index.ts",
-    "populate": "DATABASE_URL=postgres://duouser:duopassword@127.0.0.1:5432/duo env secret=qMyLZALzs229ybdQXNyzYRdju7X784TH env tokenLife=7d env baseURL=localhost:3000 env NODE_ENV=development ts-node-dev --respawn --transpileOnly ./src/populate/index.ts",
-    "generate:uows-client:dev": "uows-client-generator https://devapis.facilities.rl.ac.uk/ws/UserOfficeWebService?wsdl src/datasources/stfc/UOWSSoapInterface.ts"
+    "populate": "DATABASE_URL=postgres://duouser:duopassword@127.0.0.1:5432/duo env secret=qMyLZALzs229ybdQXNyzYRdju7X784TH env tokenLife=7d env baseURL=localhost:3000 env NODE_ENV=development ts-node-dev --respawn --transpileOnly ./src/populate/index.ts"
```
TickleThePanda, line 26: This looks a bit odd that our client generation has gone from here.

TickleThePanda (changes requested):

package.json
simonfernandes: It does, yeah! I don't know why it's gone, it wasn't intentional. I'll add it back in and try to fix the build failure

package.json
```diff
@@ -43,6 +43,7 @@
     "@esss-swap/duo-logger": "^1.0.3",
     "@esss-swap/duo-message-broker": "^1.0.2",
     "@esss-swap/duo-validation": "1.1.42",
+    "@types/sax": "^1.2.1",
```
mr-field, line 46: This should still be a dev dependency, it's not needed at runtime

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -0,0 +1,406 @@
...
+
+    return stfcUser ? toEssUser(stfcUser) : null;
+  }
+
+  async getByUsername(username: string): Promise<User | null> {
+    const stfcUser = (
+      await client.getBasicPersonDetailsFromEmail(token, username)
+    )?.return;
+
+    return stfcUser ? toEssUser(stfcUser) : null;
```
mr-field, line 173: Looks like this implmentation is the same as `getByEmail()`. Are we sure that the username is going to be an email? If we really are, we could replace this with just `return getByEmail(username)`.

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -0,0 +1,406 @@
...
+    throw new Error('Method not implemented.');
+  }
+
+  async me(id: number) {
+    return database
+      .select()
+      .from('users')
+      .where('user_id', id)
+      .first()
+      .then((user: UserRecord) => (!user ? null : createUserObject(user)));
```
mr-field, line 235: The implementation of this is the same as `get()`. Is that intentional?

mr-field (changes requested):

src/datasources/stfc/StfcUserDataSource.ts
simonfernandes: The username will actually be a usernumber when the dummy user is created, I'll change it. Thanks

src/datasources/stfc/StfcUserDataSource.ts
simonfernandes: It was, as they're the same in the Postgres data source

mr-field:
@simonfernandes  Could you try solving the merge conflict to see if that unlocks the tests?

package.json
simonfernandes: After I merged develop into my branch in https://github.com/UserOfficeProject/user-office-backend/pull/136/commits/2d83440dd681130dd6361877c5ece5df0f0ad039, the software ran fine, but the pre-commit lint failed: 

>node_modules/soap/lib/wsdl/index.d.ts(1,22): error TS7016: Could not find a declaration file for module 'sax'. 'C:/Programming/user-office-backend/node_modules/sax/lib/sax.js' implicitly has an 'any' type.
  Try `npm install @types/sax` if it exists or add a new declaration (.d.ts) file containing `declare module 'sax';`

I pushed it with `--no-verify` so that I could add sax back in commit https://github.com/UserOfficeProject/user-office-backend/pull/136/commits/4760989fd8216a220bd47f7f8c77ba2615d6509b. I can revert it but for now it doesn't seem to lint without this. Do you have any ideas?

simonfernandes:
The checks seem to be running now that I've resolved the merge conflicts, but the "Build, Test and Push" check is failing because it can't login to DockerHub. Not sure if you are aware of this @UserOfficeProject/swap

package.json
mr-field: `@types/sax` is necessary to build the application but not to run it, so it should be kept as a [dev dependency](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file#adding-dependencies-to-a-packagejson-file-from-the-command-line) like before. Does the pre-commit check fail if you remove it from the runtime dependencies and add it to the dev ones?

simonfernandes: That's done the trick, thanks! I'm not sure why it disappeared as a dev dependency

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -0,0 +1,309 @@
...
+    undefined,
+    stfcUser.familyName ?? '',
+    stfcUser.email ?? '',
+    stfcUser.firstNameKnownAs ?? '',
+    '',
+    '',
+    '',
+    0,
+    '',
+    0,
```
mr-field, line 64: Hardcoding these values will give every user the same nationality and organisation; is it going to be updated as part of https://github.com/UserOfficeProject/stfc-user-office-project/issues/46?

TickleThePanda (approved):

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -0,0 +1,309 @@
...
+    throw new Error('Method not implemented.');
+  }
+
+  async getUserRoles(id: number): Promise<Role[]> {
+    const stfcRoles: stfcRole[] | null = (
+      await client.getRolesForUser(token, id)
+    )?.return;
+
+    if (!stfcRoles) {
+      return [new Role(0, Roles.USER, 'User')];
```
mr-field, line 175: It seems that we're hardcoding the role ID as `0` here (and lower down the same happens with `2` and `7`). Can we actually rely on these IDs remaining fixed in the future? If not, a more stable solution might be to get all roles with `getRoles()`, then selecting the right one with something like `allRoles.find(role => x.shortCode = Roles.USER)`

mr-field (changes requested):

src/datasources/stfc/__mocks__/UOWSSoapInterface.ts
```diff
@@ -11,4 +11,36 @@ export default class UOWSSoapClient {
       },
     };
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public getRolesForUser(token: any, userNumber: any): any {
```
martin-trajanovski, line 16: Do we really need `token` and `userNumber` here because I do not see any usage of it. If we do need them maybe we you can add the correct types instead of just using `any` and remove the eslint disable comment on top after defining the types. I see we have the same thing in the previous function as well `getPersonDetailsFromSessionId` where you can try to specify the types.

martin-trajanovski (approved):

src/datasources/stfc/StfcUserDataSource.ts
simonfernandes: That sounds like a good reason to keep https://github.com/UserOfficeProject/stfc-user-office-project/issues/46 open, as there isn't a completely straightforward solution for the nationalities at least.

Does that sound ok to you and @TickleThePanda? I'll add some more detail to that issue if so.

src/datasources/stfc/StfcUserDataSource.ts
simonfernandes: What I found was that the ID actually had no impact, so I just copied their current database order. But your solution is a really nice one so I'll update this

src/datasources/stfc/StfcUserDataSource.ts
TickleThePanda: Do you mind splitting UserOfficeProject/stfc-user-office-project#46 into two parts: one to get basic user details; one to get organisation and nationality, please?

Split meaning close and add the label "triage: duplicate".

This can then close the basic user details side of things.

src/datasources/stfc/StfcUserDataSource.ts
simonfernandes: This turned out to be a larger change than I had expected but I think this is what we want - to only grant a role if the definition exists in the database. This mimicks how it works on the ESS side. This is in commit https://github.com/UserOfficeProject/user-office-backend/pull/136/commits/9f79d3a428f5b9a4796aa3a0ddddf3649138829b, please ignore https://github.com/UserOfficeProject/user-office-backend/pull/136/commits/d63e27719caa0d889bf92e1d315d9b3432fb06ea as I meant to overwrite that

src/datasources/stfc/__mocks__/UOWSSoapInterface.ts
simonfernandes: I didn't think this would work without the parameters but it does, thanks

mr-field (approved):

PR #150, opened 2021-02-01
SWAP-1320 When trying to delete a proposal with dependencies one get thrown out

peter-asztalos:
## Description

This PR fixes an unhandled error that occurs when we trying to delete a proposal that has other rows in the DB that depend on it. 

## How Has This Been Tested

- [x] e2e test
- [x] manual test 

## Fixes

https://jira.esss.lu.se/browse/SWAP-1320

## Changes

- check if the error is a `foreign_key_violation`

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/219


martin-trajanovski (approved):

src/mutations/ProposalMutations.ts
```diff
@@ -198,11 +199,21 @@ export default class ProposalMutations {
...
 
-    return result || rejection('INTERNAL_ERROR');
+      return result;
+    } catch (e) {
+      if ('code' in e && e.code === '23503') {
+        return rejection(
+          `Failed to delete proposal with ID "${proposal.shortCode}", it has dependencies which need to be deleted first` as ResourceId
+        );
+      }
+
```
jekabs-karklins, line 214: Maybe we should use a logger and log it?

peter-asztalos: Not sure if it's relevant for logging. It's only interesting for the user who tries to delete a proposal but the action fails because the proposal has related data, it's on the user to decide what to do from here (delete the dependency or keep the proposal).

jekabs-karklins: No, I mean in the rest of the cases where e.code !== '23503'

peter-asztalos: Ahh, I see, I'll add one for that, although the `query-error` listener in the `database.ts` should already log any db related errors

PR #153, opened 2021-02-02
SWAP-1319 Cloning the sample the topics are incomplete

peter-asztalos:
## Description

This PR fixes sample cloning in questionaries by copying completed topics also.

## How Has This Been Tested

- [x] e2e test
- [x] manual test

## Fixes

https://jira.esss.lu.se/browse/SWAP-1319

## Changes

- copy completed topics
- fix possible null crash
- wait for user rights lookup to resolve
- sort samples by `created_at` so it fixes the jumping and reordering behavior on FE.

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/222

### DIFF
--- a/src/datasources/postgres/SampleDataSource.ts
+++ b/src/datasources/postgres/SampleDataSource.ts
@@ -77,6 +77,10 @@ export default class PostgresSampleDataSource implements SampleDataSource {
       .where('sample_id', sampleId)
       .first()
       .then((sample?: SampleRecord) => {
+        if (!sample) {
+          logger.logError('Sample does not exist', { sampleId });
+        }
+
         return sample ? createSampleObject(sample) : null;
       });
   }
--- a/src/datasources/postgres/SampleDataSource.ts
+++ b/src/datasources/postgres/SampleDataSource.ts
@@ -71,16 +71,13 @@ export default class PostgresSampleDataSource implements SampleDataSource {
       });
   }
 
-  getSample(sampleId: number): Promise<Sample> {
+  getSample(sampleId: number): Promise<Sample | null> {
     return database('samples')
       .select('*')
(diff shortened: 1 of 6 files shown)


martin-trajanovski (approved):

src/datasources/postgres/SampleDataSource.ts
```diff
@@ -71,16 +71,13 @@ export default class PostgresSampleDataSource implements SampleDataSource {
...
   }
 
-  getSample(sampleId: number): Promise<Sample> {
+  getSample(sampleId: number): Promise<Sample | null> {
     return database('samples')
       .select('*')
       .where('sample_id', sampleId)
-      .then((records: SampleRecord[]) => {
-        if (records.length !== 1) {
-          logger.logError('Sample does not exist', { sampleId });
```
jekabs-karklins, line 80: I think it would be interesting to see logs when someone tries to access sample that does not exist, as it should not normally happen.

jekabs-karklins, line 80: I think it would be interesting to see logs when someone tries to access sample that does not exist, as it should not normally happen.

src/mutations/SampleMutations.ts
```diff
@@ -122,12 +122,17 @@ export default class SampleMutations {
     if (!agent) {
       return rejection('NOT_AUTHORIZED');
     }
-    if (!this.sampleAuthorization.hasWriteRights(agent, sampleId)) {
+    if (!(await this.sampleAuthorization.hasWriteRights(agent, sampleId))) {
```
jekabs-karklins, line 125: :+1:

jekabs-karklins (approved):

PR #156, opened 2021-02-03
Swap 1283 api access

martin-trajanovski:
## Description

Functionality that allows user officers to create/update/delete API access tokens with specific permissions for external services to be able to read/modify data.  

## Motivation and Context

This change is because of a need for multiple external services to mainly read but also modify data in user office.

## How Has This Been Tested

- Unit tests
- e2e tests
- Manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-1283
https://jira.esss.lu.se/browse/SWAP-1338

## Changes

New table added in the database called `api_permissions` where we store the tokens and permissions for the tokens. Queries and mutations to support reading and writing to the table and unit tests.


src/utils/helperFunctions.ts
```diff
@@ -30,3 +30,12 @@ export const mergeValidationSchemas = (...schemas: Yup.ObjectSchema[]) => {
...
   return merged;
 };
+
+export const generateUniqueId = () => {
+  return (
+    Date.now().toString(36) +
+    Math.random()
+      .toString(36)
+      .substring(2)
+  );
```
peter-asztalos, line 40: Although it looks random, but to follow good practices I'd recommend `crypto.randomBytes(...)`, like:

```typescript

import {randomBytes} from 'crypto';

// ....

const numberOfBytes = 16; // e. g.

// ...
randomBytes(numberOfBytes).toString('hex') // or can be 'base64'
// this results in hex: c74fe8cf8838712da6cddccd31cbf06a
// or with base64: o+NgMJRlT85N7zrAdN3j6Q== (if numberOfBytes is multiple of 6 then it will drop the padding (the = sign))

```

src/mutations/AdminMutations.ts
```diff
@@ -99,4 +110,67 @@ export default class AdminMutations {
...
+
+  @ValidateArgs(createApiAccessTokenValidationSchema(IS_BACKEND_VALIDATION))
+  @Authorized([Roles.USER_OFFICER])
+  async createApiAccessToken(
+    agent: UserWithRole | null,
+    args: CreateApiAccessTokenInput
+  ) {
+    const accessTokenId = generateUniqueId();
+    const accessPermissions = JSON.parse(args.accessPermissions);
+    const generatedAccessToken = `Bearer ${signToken({ accessTokenId })}`;
```
peter-asztalos, line 122: I think we don't need `Bearer` part, it's not part of the token.

src/queries/AdminQueries.ts
```diff
@@ -25,4 +29,70 @@ export default class AdminQueries {
...
+  @Authorized([Roles.USER_OFFICER])
+  async getAllQueryAndMutationMethods(
+    agent: UserWithRole | null,
+    context: BasicResolverContext
+  ) {
+    const allQueryMethods: string[] = [];
+    const allMutationMethods: string[] = [];
+
+    for (const queryKey in context.queries) {
+      if (Object.prototype.hasOwnProperty.call(context.queries, queryKey)) {
```
peter-asztalos, line 59: I think this part (the `for` and the `if`) can be simplified with `Object.keys(...).forEach(...)`
Might also be true for the following lines (`Object.getPrototypeOf(...)`, `Object.getOwnPropertyNames(...)`)

peter-asztalos (approved):

src/datasources/postgres/AdminDataSource.ts
```diff
@@ -229,12 +234,123 @@ export default class PostgresAdminDataSource implements AdminDataSource {
     });
   }
 
-  getFeatures(): Promise<Feature[]> {
+  async getFeatures(): Promise<Feature[]> {
```
jekabs-karklins, line 237: Do we need to have async here or is it just for consistency?

jekabs-karklins (approved):

src/datasources/postgres/AdminDataSource.ts
martin-trajanovski: Yes just for consistency to be the same as every method in the class.

PR #160, opened 2021-02-04
SWAP-1251: Add table for units with delete,get and create endpoints

bolmsten:
## Description

As a user officer I can now add units in the GUI instead of code

## Motivation and Context

The possibility of adding units in GUI will reduce the need for development effort each time a unit is added

## Changes

Adds a new table in the database called units and associated endpoints for creation and deletion


src/datasources/postgres/AdminDataSource.ts
```diff
@@ -19,9 +20,59 @@ import {
...
+      throw new Error(`Could not delete institution with id:${id}`);
+    }
+
+    return {
+      id: unitRecord.unit_id,
+      name: unitRecord.unit,
+    };
+  }
+  async getUnits(): Promise<Unit[]> {
+    return database
```
martin-trajanovski, line 62: I see `await` is used in the previous two functions and it could be used here as well to keep the consistency. And not sure why we need to use two `orderBy` methods.

src/mutations/AdminMutations.ts
```diff
@@ -79,6 +81,19 @@ export default class AdminMutations {
...
+  @Authorized([Roles.USER_OFFICER])
+  async createUnit(agent: UserWithRole | null, args: CreateUnitArgs) {
+    const unit = new Unit(0, args.name);
+
+    return await this.dataSource.createUnit(unit);
+  }
+
+  @Authorized([Roles.USER_OFFICER])
+  async deleteUnit(agent: UserWithRole | null, id: number) {
+    //Do checks here to not delete a unit used
```
martin-trajanovski, line 93: Isn't this required from the beginning (to do the checks if unit is used) or it can be done later?

martin-trajanovski (changes requested):

src/datasources/postgres/AdminDataSource.ts
```diff
@@ -19,9 +20,59 @@ import {
...
+  async createUnit(unit: Unit): Promise<Unit | null> {
+    const [unitRecord]: UnitRecord[] = await database
+      .insert({
+        unit: unit.name,
+      })
+      .into('units')
+      .returning('*');
+
+    if (!unitRecord) {
+      throw new Error('Could not create call');
```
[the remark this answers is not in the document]
bolmsten: changed

src/datasources/postgres/AdminDataSource.ts
```diff
@@ -19,9 +20,59 @@ import {
...
+
+  async deleteUnit(id: number): Promise<Unit> {
+    const [unitRecord]: UnitRecord[] = await database('units')
+      .where('units.unit_id', id)
+      .del()
+      .from('units')
+      .returning('*');
+
+    if (!unitRecord) {
+      throw new Error(`Could not delete institution with id:${id}`);
```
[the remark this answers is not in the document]
bolmsten: changed

src/mutations/AdminMutations.ts
bolmsten: It does not need to check as the answer and options is stored with the question.

martin-trajanovski (approved):

db_patches/0078_AddUnits.sql
```diff
@@ -0,0 +1,21 @@
...
+BEGIN
+	IF register_patch('AddUnits.sql', 'fredrikbolmsten', 'Adding units', '2021-02-02') THEN
+	BEGIN
+
+      CREATE TABLE units (
+            unit_id serial UNIQUE,
+            unit varchar(50) DEFAULT NULL
+      ); 
+
+    INSERT INTO units (unit) values ('kelvin');
```
jekabs-karklins, line 14: Will the remaining units be added through the UI?

jekabs-karklins (approved):
Looks great

db_patches/0078_AddUnits.sql
bolmsten: yes, but added these for the E2E test

PR #174, opened 2021-02-15
Swap 1381 call proposal count

martin-trajanovski:
## Description

Adding `proposalCount` resolver on the call.

## Motivation and Context

To be able to see how many proposals are on the call.

## How Has This Been Tested

e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1381

### DIFF
--- a/src/resolvers/types/Call.ts
+++ b/src/resolvers/types/Call.ts
@@ -81,6 +81,11 @@ export class CallInstrumentsResolver {
       call.proposalWorkflowId
     );
   }
+
+  @FieldResolver(() => Int, { nullable: true })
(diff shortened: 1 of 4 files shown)


src/resolvers/types/Call.ts
```diff
@@ -81,6 +81,11 @@ export class CallInstrumentsResolver {
       call.proposalWorkflowId
     );
   }
+
+  @FieldResolver(() => Int, { nullable: true })
```
peter-asztalos, line 85: I think this can't be nullable

peter-asztalos (approved):

PR #177, opened 2021-02-17
Swap 1348 update status proposal table

martin-trajanovski:
## Description

Returning the next proposal status if exists on `assignProposalToSEP` mutation, so we can update the frontend proposals table immediately if it is successful. 

## Motivation and Context

The proposal officer table in the frontend wasn't updated at real time when there was proposal workflow status change.

## How Has This Been Tested

- manual testing
- e2e tests on the frontend

## Fixes

https://jira.esss.lu.se/browse/SWAP-1348

## Changes

Added function to grab the next status in proposal workflow by proposal id.


src/datasources/postgres/SEPDataSource.ts
```diff
@@ -477,12 +480,42 @@ export default class PostgresSEPDataSource implements SEPDataSource {
...
+      .join('next_status_events as nse', {
+        'nse.proposal_workflow_connection_id':
+          'pwc.proposal_workflow_connection_id',
+      })
+      .where('proposal_id', proposalId)
+      .andWhere('nse.next_status_event', event)
+      .first();
+
+    return new ProposalStatus(
+      nextProposalStatus?.proposal_status_id,
```
peter-asztalos, line 511: I think returning null if `nextProposalStatus` not found at all would be more cleaner.

peter-asztalos (approved):
Looks good (☞ﾟヮﾟ)☞

PR #181, opened 2021-02-22
Swap 1395: Add delete for Calls

bolmsten:
## Description

A user officer should be able to delete a call if there is not proposals attached to it

## Motivation and Context

It can be that calls have been made by mistake or all proposals have been moved to another call. It should than be possible to delete a call

## How Has This Been Tested

Unit test

## Fixes

SWAP-1395


martin-trajanovski (approved):
Looks good!

PR #182, opened 2021-02-23
Hot fix for SEP assigment

bolmsten:
## Description

At the moment a proposal can have multiple SEP's assigned to it, this will cause the proposal to appear multiple times in the list

## Motivation and Context

See above

### DIFF
--- a/src/mutations/SEPMutations.ts
+++ b/src/mutations/SEPMutations.ts
@@ -193,7 +193,7 @@ export default class SEPMutations {
     args: AssignProposalToSEPArgs
   ): Promise<ProposalIdsWithNextStatus | Rejection> {
     const SEP = await this.dataSource.getSEPByProposalId(args.proposalId);
-    if (SEP !== null) {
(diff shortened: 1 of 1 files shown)


src/mutations/SEPMutations.ts
```diff
@@ -192,6 +192,20 @@ export default class SEPMutations {
     agent: UserWithRole | null,
     args: AssignProposalToSEPArgs
   ): Promise<ProposalIdsWithNextStatus | Rejection> {
+    const SEP = await this.dataSource.getSEPByProposalId(args.proposalId);
+    if (SEP !== null) {
```
martin-trajanovski, line 196: Can we just use like `if (SEP)` instead of `if (SEP !== null)` that would be cleaner and I guess will check as well for values like `undefined` and so on.

martin-trajanovski (changes requested):

martin-trajanovski (approved):

PR #188, opened 2021-02-24
Revert "Hot fix for SEP assigment"

peter-asztalos:
Reverts UserOfficeProject/user-office-backend#182


martin-trajanovski (approved):

martin-trajanovski:
I see this reverts #182 but do you plan to add that change again or?

peter-asztalos:
We should but with updated e2e tests.

PR #201, opened 2021-03-04
fix reviews filter

peter-asztalos:
fix reviews filter

### DIFF
--- a/src/datasources/postgres/ReviewDataSource.ts
+++ b/src/datasources/postgres/ReviewDataSource.ts
@@ -187,21 +187,23 @@ export default class PostgresReviewDataSource implements ReviewDataSource {
       .select()
       .from('SEP_Reviews')
       .modify((qb) => {
-        if (callId !== undefined) {
+        // sometimes the ID 0 is sent as a equivalent of all
+        if (callId) {
           qb.join('proposals', {
             'proposals.proposal_id': 'SEP_Reviews.proposal_id',
           });
           qb.where('proposals.call_id', callId);
         }
 
-        if (instrumentId !== undefined) {
+        // sometimes the ID 0 is sent as a equivalent of all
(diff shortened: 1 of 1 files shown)


src/datasources/postgres/ReviewDataSource.ts
```diff
@@ -187,21 +187,23 @@ export default class PostgresReviewDataSource implements ReviewDataSource {
...
+        // sometimes the ID 0 is sent as a equivalent of all
+        if (instrumentId) {
           qb.join('instrument_has_proposals', {
             'instrument_has_proposals.proposal_id': 'SEP_Reviews.proposal_id',
           });
           qb.where('instrument_has_proposals.instrument_id', instrumentId);
         }
 
-        if (status !== undefined) {
+        if (status !== undefined && status !== null) {
```
martin-trajanovski, line 206: how can this be `true` do not understand. Maybe you need `| |` operator here

martin-trajanovski (changes requested):

src/datasources/postgres/ReviewDataSource.ts
martin-trajanovski: Ah I see this is `!==` I thought it is `===` 😄

martin-trajanovski (approved):

jekabs-karklins (approved):

src/datasources/postgres/ReviewDataSource.ts
peter-asztalos: No, it's intended, although the types suggest undefined or ReviewStatus, nullable graphql fields can be null actually: you can leave out (undefined) or use null as a value that's why I'm checking for both and only accepting if the status is neither of them. This is something that we may want to fix everywhere.

PR #209, opened 2021-03-09
SWAP-1306 As a user I want to see when my proposal has been scheduled

peter-asztalos:
## Description

This PR adds support for Users to see their upcoming (or all) beam times.

## How Has This Been Tested

- [x] manual
- [x] e2e test 

## Fixes

https://jira.esss.lu.se/browse/SWAP-1306

## Changes

- support to access scheduler
- core and scheduler github actions updated
- core and scheduler docker compose files updated
- API gateway updated


martin-trajanovski (approved):
Looks good! 💯

example.env
```diff
@@ -14,6 +14,8 @@ EAM_AUTH_SECRET=secret
 EAM_AUTH_USER=user
 EAM_AUTH_PASS=pass
 EAM_API_URL=https://ios.esss.lu.se:7443
+# Enable to disallow EAM
+UO_FEATURE_DISABLE_EAM=1
```
jekabs-karklins, line 18: +1

example.env
```diff
@@ -34,5 +36,5 @@ NODE_ENV=development
 DATABASE_URL=postgres://duouser:duopassword@127.0.0.1:5432/duo
 baseURL=localhost:3000
 tokenLife=7d
-secret=secret
+secret=qMyLZALzs229ybdQXNyzYRdju7X784TH
```
jekabs-karklins, line 39: It's best practice not to place any passwords or secrets into the repository, even if they are not working/dangerous. This is because secrets like this encourage curiosity for poking arround.

jekabs-karklins (changes requested):

example.env
peter-asztalos: I totally agree with you, but in this case, the secret is already part of the git history and is already part of other files. As this secret is used by the token we use for e2e testing, I think the best is to have it here as a default value. Ofc. we have to make sure we don't use it anywhere else (testing, prod).

docker-compose.yml
```diff
@@ -1,34 +1,21 @@
...
-  reverse-proxy:
+  # used to proxy the scheduler UI FE and BE requests to the right container
+  # Scheduler FE is accessible on localhost:33000/
+  # API gateway is accessible on localhost:33000/gateway
+  scheduler-proxy:
     image: traefik:1.7 # The official Traefik docker image
     command: --api --docker # Enables the web UI and tells Traefik to listen to docker
     ports:
-      - "8081:80" # The HTTP port
+      - "33000:80" # The HTTP port
```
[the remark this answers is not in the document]
peter-asztalos: Previously this part was used to bootstrap a totally functioning UO, including already deployed docker images (if we want to have things like this we should use a different docker-compose file), but this now used for accessing the scheduler UI. Since the port we use on FE is 3000 for development, I thought having a similar one is better, and there is less chance we would have a port conflict.

docker-compose.yml
```diff
@@ -40,27 +27,74 @@ services:
...
+      DATABASE_USER: duouser
+      DATABASE_PASSWORD: duopassword
+      DATABASE_DATABASE: duo
     ports:
       - 4500:4500
     cap_add:
       - SYS_ADMIN
+
+  gateway:
+    image: "dmsc/duo-gateway:SWAP-1306"
```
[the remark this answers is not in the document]
peter-asztalos: only for testing, will fix it before merge

Scott-James-Hurley (approved):

jekabs-karklins (approved):

PR #211, opened 2021-03-09
SWAP-1410 #resolved

martin-trajanovski:
## Description

Reset proposal events that are coming after given status. For example if proposal had SEP_REVIEW status and we manually reset to FEASIBILITY_REVIEW then events like: proposal_feasible, proposal_feasibility_review_submitted and proposal_sep_selected should be reset to false in the proposal_events table

## Motivation and Context

Previously if we manually reset proposal status the events were staying in the state they are nothing was changed there to reset them at the point where proposal is returned.

## How Has This Been Tested

manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1410

## Changes

Added new method for reseting proposal events inside proposal data source.

### DIFF
--- a/src/datasources/ProposalDataSource.ts
+++ b/src/datasources/ProposalDataSource.ts
@@ -47,4 +47,9 @@ export interface ProposalDataSource {
     callId: number,
     templateId: number
   ): Promise<Proposal>;
+  resetProposalEvents(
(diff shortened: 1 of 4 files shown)


jekabs-karklins (approved):
Looks good

PR #217, opened 2021-03-15
#21: UpdatesStatusTableDBPatch

davidsymons1999:
## Description

This is the change made to the backend for the issue https://github.com/UserOfficeProject/stfc-user-office-project/issues/80 the changes update the `AddProposalStatusesTable` DB patch.

## Motivation and Context

This change allows users to edit proposals after submission, while also retaining the ability to continue with the standard proposal workflow.


db_patches/0059_AddProposalStatusesTable.sql
```diff
@@ -20,6 +20,7 @@ BEGIN
         INSERT INTO proposal_statuses (name, description) VALUES ('NOT_ALLOCATED', 'Proposal is not allocated.');
         INSERT INTO proposal_statuses (name, description) VALUES ('SCHEDULING', 'Proposal should be scheduled.');
         INSERT INTO proposal_statuses (name, description) VALUES ('EXPIRED', 'Proposal has expired.');
+        INSERT INTO proposal_statuses (name, description) VALUES ('EDITABLE_SUBMISSION', 'Proposal is editable after submission.');
```
martin-trajanovski, line 23: I think this wont run because it is already been executed. If you want to add something new you need to do it in a new patch.  You should create new file at the end of patches and add this line into it. We have procedure that executes the patches that re not executed only.

davidsymons1999:
Would it be a good idea to check the DB patch numbers once a month or so? Currently, there are multiples of the patches starting from `0060_` or is the idea that the numbers are updated by those that notice? If so I can do that now?

db_patches/0078_Insert_Into_Proposal_Statuses_Table.sql
```diff
@@ -0,0 +1,14 @@
+
+
+DO
+$$
+BEGIN
+	IF register_patch('Insert_Into_Proposal_Statuses_Table.sql', 'David Symons', 'Add additional Status into Proposal Statuses table', '2021-03-19') THEN
+    BEGIN
+        
+        INSERT INTO proposal_statuses (name, description) VALUES ('EDITABLE_SUBMISSION', 'Proposal is editable after submission.');
```
mr-field, line 9: I'd prefer to call the state `EDITABLE_SUBMITTED`, which better matches the existing ones

db_patches/0059_AddProposalStatusesTable.sql
mr-field: Now that we have a new patch for the adding the state, this should be removed

mr-field (commented):
> Currently, there are multiple duplicates of the patch numbers starting from 0060_ or is the idea that the numbers are updated by those that notice?

I think it's fine to have multiple patches starting with the same number, as long as they have different names. I suspect that patches with the same number are coupled and were introduced together.

jekabs-karklins (approved):

mr-field (approved):

PR #235, opened 2021-03-29
Works

jekabs-karklins:


src/config/dependencyConfigE2E.ts
```diff
@@ -0,0 +1,51 @@
...
+mapClass(Tokens.ShipmentDataSource, PostgresShipmentDataSource);
+mapClass(Tokens.SystemDataSource, PostgreSystemDataSource);
+mapClass(Tokens.TemplateDataSource, PostgresTemplateDataSource);
+mapClass(Tokens.UserDataSource, PostgresUserDataSource);
+
+mapClass(Tokens.AssetRegistrar, SkipAssetRegistrar);
+
+mapClass(Tokens.MailService, SkipSendMailService);
+
+mapValue(Tokens.PostToMessageQueue, createSkipPostingHandler);
```
peter-asztalos, line 51: Here it's a function reference, but in other config files, it's called.
So it should be either `createSkipPostingHandler` or `createSkipPostingHandler()`

src/eventHandlers/email.ts
```diff
@@ -1,35 +1,23 @@
...
-const mailService: MailService =
-  process.env.EMAIL_PROTOCOL === 'SMTP'
-    ? new SMTPMailService()
-    : new SparkPostMailService(options);
-
-const isDevEnv = process.env.NODE_ENV === 'development';
-
-// in dev env don't try to send email
-if (isDevEnv) {
-  mailService.sendMail = async (...args: any[]): Promise<any> => 'no-op';
```
peter-asztalos, line 26: I think we should replicate this logic somehow in the `SparkPostMailService` or somewhere else, so when we working locally, deploy it to user office testing we won't get errors constantly that spark is not configured.

src/mutations/ShipmentMutations.ts
```diff
@@ -88,11 +100,11 @@ export default class ShipmentMutations {
   }
 
   async submitShipment(agent: UserWithRole | null, args: SubmitShipmentArgs) {
-    if (!this.shipmentAuthorization.hasWriteRights(agent, args.shipmentId)) {
+    if (!this.shipmentAuth.hasWriteRights(agent, args.shipmentId)) {
```
peter-asztalos, line 103: This and other `hasWriteRights` uses should include `await`

src/workflowEngine/index.ts
```diff
@@ -40,10 +48,7 @@ const updateProposalStatus = (
   proposalId: number,
   nextProposalStatusId: number
 ) => {
-  return proposalDataSource.updateProposalStatus(
-    proposalId,
-    nextProposalStatusId
-  );
+  proposalDataSource.updateProposalStatus(proposalId, nextProposalStatusId);
```
peter-asztalos, line 51: `return` statement is missing

src/mutations/ShipmentMutations.ts
peter-asztalos: `src/mutations/SampleMutations.ts` has the same issues

peter-asztalos (changes requested):
Nice job, I very like this approach, separating things in config like that also makes things easier.

simonfernandes (commented):
I can't really comment on the implementation details of this, but it looks like a nice change that will be very useful, and the STFC config also looks fine 👍

src/eventHandlers/email.ts
jekabs-karklins: I actually pondered about this a little. I think we should keep the local development environment as well as user office test as close to production as possible, to avoid bugs.

I don't know what is the right way to solving the errors about Sparkpost not being configured in the test environments is though.

bolmsten (approved):

src/config/dependencyConfigE2E.ts
jekabs-karklins: It should be a function call, thanks

src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -17,8 +20,12 @@ import {
   InstrumentHasProposalsRecord,
 } from './records';
 
+@injectable()
```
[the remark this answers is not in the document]
jekabs-karklins: @singleton decorators have no meaning in this case, so I removed them. To register as a singleton it needs to be done when registering for the class. 
If interested look at https://github.com/microsoft/tsyringe/issues/158#issuecomment-810874547
Or mapClass() function on our project

src/mutations/ShipmentMutations.ts
jekabs-karklins: Done and I also added tests to cover all places. Thanks

src/workflowEngine/index.ts
jekabs-karklins: +1

peter-asztalos (approved):

PR #237, opened 2021-03-30
Reference numbers for submitted proposals

simonfernandes:
## Description
This adds reference numbers for proposals. The idea is that a call can optionally be given a reference number format, where a proposal submitted in that call will then be given a reference number following that format. The reference number itself will overwrite the existing proposal short code. Reference numbers of proposals will also be updated when adding or modifying a format on the call. If a call doesn't have a reference number format, the short code of a proposal isn't changed on submission.

I have used Postgres features for handling any potential concurrency issues and it seems to work nicely. It does mean that there is a bit more logic in the datasource layer than perhaps would be ideal, but queries have to use the "for update" syntax and be done in a single transaction, and the mutation layer doesn't/shouldn't have any knowledge of a database. I did have a read up on transaction isolation levels and I think that setting the level to serializable isn't necessary. 

I have had to add some new database columns:

**call.reference_number_format**
This holds what the format should be for all submitted proposals in the call.

**call.proposal_sequence**
This value is the sequence number that the next submitted proposal should use, and it is updated regardless of whether the call has a format. It is necessary because deleted proposals should still retain their sequence in the call. E.g. if the 65th proposal in a call gets deleted, the proposal_sequence knows that the next proposal in the sequence should be 66. This could maybe be done with a count query that includes deleted, though it depends on whether rows ever get completely removed.

**proposals.reference_number_sequence**
This is the proposal's order in the call's sequence of submitted proposals. It is updated even if the call doesn't have a format, and is used when modifying a call's format. If the format is changed, all proposals in the call should be updated and retain their original order. The only way to know which number each proposal is in the call is by looking at this value. There is perhaps another way to do this, by reverse engineering the original reference number, but that starts to get messy and also doesn't work with the scenario in which a reference number format is added to a call that didn't have one originally.

## Motivation and Context

It is an STFC requirement - see https://github.com/UserOfficeProject/stfc-user-office-project/issues/5 for more info

## How Has This Been Tested
I have written unit tests and integration tests, and modified an existing e2e test.

The existing package.json scripts for the unit tests have been modified to point at a new jest config - this config looks for `spec.ts` files. 

The integration tests use a separate jest config that looks for `test.ts` files. I have made an attempt at getting these to run in a GitHub action by largely copying the e2e config, I haven't been able to test this though.

## Fixes

Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/5

## Changes

There are changes to how a proposal is submitted, and how a call is updated.

## Depends on

There is a frontend branch, I'm having trouble regenerating the SDK right now though


bolmsten (approved):

jest.unit.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  roots: ['<rootDir>/src'],
+  transform: {
+    '^.+\\.tsx?$': 'ts-jest',
+  },
+  testEnvironment: 'node',
+  testRegex: '(/__tests__/.*|(\\.|/)(spec))\\.[jt]sx?$',
+  collectCoverage: true,
```
jekabs-karklins, line 8: Try adding here
 `setupFilesAfterEnv: ['<rootDir>/src/config/dependencyConfigTest.ts'],`

jekabs-karklins (approved):

PR #239, opened 2021-03-31
SWAP-1492 Scheduler module should pick up a proposal when it goes into the "scheduled" state

peter-asztalos:
## Description

This PR updates how proposal booking create/update events are created.

## Motivation and Context

The current logic is a bit old and has no reliable way to retry if something goes wrong or update values if the proposal changed.

## How Has This Been Tested

- [x] existing e2e test passing
- [x] manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-1492

## Changes

- yaml files formatting fixed
- new events added that are triggered when proposal status changes (by the user or by workflow engine)
- message broker emitter updated to use these new events to trigger proposal booking creation/update
- unused gitlab-cy.yml file removed

## Depends on

https://github.com/UserOfficeProject/user-office-scheduler-backend/pull/25
https://github.com/UserOfficeProject/user-office-scheduler-frontend/pull/26
https://github.com/UserOfficeProject/user-office-frontend/pull/307
https://github.com/UserOfficeProject/user-office-lib/pull/34


src/eventHandlers/messageBroker.ts
```diff
@@ -39,43 +41,49 @@ export default function createHandler({
     }
 
     switch (event.type) {
-      // case Event.PROPOSAL_ACCEPTED: {
```
martin-trajanovski, line 42: I guess you will cleanup the comments later

src/mutations/ProposalMutations.ts
```diff
@@ -242,6 +242,7 @@ export default class ProposalMutations {
   }
 
   @EventBus(Event.PROPOSAL_MANAGEMENT_DECISION_UPDATED)
+  @EventBus(Event.PROPOSAL_STATUS_CHANGED_BY_USER)
```
martin-trajanovski, line 245: We don't need this event to be fired here. The status will be removed from the admin form and status can be changed only by separated mutation `changeProposalsStatus`. That fires already `PROPOSAL_STATUS_UPDATED` and I think that is good because we are catching that event in the workflow and firing `PROPOSAL_STATUS_CHANGED_BY_USER`.

martin-trajanovski (changes requested):

src/mutations/ProposalMutations.ts
peter-asztalos: I see, that's definitely cleaner.
But, unless it is already in develop, I think we still need this part, after the status is removed from the admin form we can delete this line, otherwise it may be confusing why the changes aren't picked up.

So this line should be removed after those changes, will see which one gets in first.

src/eventHandlers/messageBroker.ts
peter-asztalos: Uhm, these are already deleted

martin-trajanovski: Oh yes I see now.

martin-trajanovski (approved):

PR #243, opened 2021-04-01
Swap 1479 fixing proposal workflow bugs

martin-trajanovski:
## Description

Fixed couple of bugs when removing and inserting connections in the workflows. Also fixed some small workflow engine issues.

## How Has This Been Tested

e2e tests
manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1479


src/datasources/mockups/ProposalSettingsDataSource.ts
```diff
@@ -146,7 +146,14 @@ export class ProposalSettingsDataSourceMock
 
   async getProposalWorkflowConnectionsById(
     proposalWorkflowId: number,
-    proposalWorkflowConnectionId: number
+    proposalWorkflowConnectionId: number,
+    {
+      nextProposalStatusId,
+      prevProposalStatusId,
+    }: {
```
jekabs-karklins, line 153: I see you use this type several times. Maybe it makes sense to factor it out into separate file and then reuse the interface

src/mutations/InstrumentMutations.ts
```diff
@@ -267,7 +267,7 @@ export default class InstrumentMutations {
     args: InstrumentSubmitArgs
   ): Promise<InstrumentHasProposals | Rejection> {
     if (
-      !(await this.userAuth.isUserOfficer(agent)) &&
+      !this.userAuth.isUserOfficer(agent) &&
```
jekabs-karklins, line 270: aren't you missing `await` here?

jekabs-karklins (changes requested):
Looks good except the missing `await`

src/mutations/InstrumentMutations.ts
martin-trajanovski: This function isUserOfficer is not an async function. It just checks if that person has that specific role in current role.

jekabs-karklins (approved):

PR #255, opened 2021-04-09
SWAP-1527 Bugfixes and improvements for sample declaration

jekabs-karklins:
## Description

This PR fixes the issue that samples were not cloned when the proposal was cloned.

## Motivation and Context

This PR also includes a refactoring that was required to complete the task. Previously sample connection to the proposal was kept in Answers table and in Samples table. This duplication has now been eliminated and is much cleaned. Information is only stored in the Samples table

## How Has This Been Tested

- [x] e2e

## Fixes

SWAP-1527 

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/321


src/mutations/ProposalMutations.ts
```diff
@@ -398,13 +409,87 @@ export default class ProposalMutations {
...
+        call
+      );
+
+      const clonedQuestionary = await this.questionaryDataSource.clone(
+        sourceProposal.questionaryId
+      );
+
+      // if user clones the proposal then it is his/her,
+      // but if userofficer, then it will belong to original proposer
+      const proposerId = (await this.userAuth.isUserOfficer(agent))
```
martin-trajanovski, line 440: I think there is no need for await here. `isUserOfficer` is not an async function.

src/mutations/ProposalMutations.ts
```diff
@@ -398,13 +409,87 @@ export default class ProposalMutations {
...
+      await this.proposalDataSource.setProposalUsers(
+        clonedProposal.id,
+        proposalUsers.map((user) => user.id)
+      );
+
+      const proposalSamples = await this.sampleDataSource.getSamples({
+        filter: { proposalId: sourceProposal.id },
       });
+
+      for await (const sample of proposalSamples) {
```
martin-trajanovski, line 476: I think it is better to use something like `Promise.all` but it doesn't look bad even like this.

martin-trajanovski (approved):
Generally it looks good!

src/mutations/ProposalMutations.ts
jekabs-karklins: ✔

src/mutations/ProposalMutations.ts
```diff
@@ -399,13 +410,87 @@ export default class ProposalMutations {
...
+        sourceProposal.questionaryId
+      );
+
+      // if user clones the proposal then it is his/her,
+      // but if userofficer, then it will belong to original proposer
+      const proposerId = this.userAuth.isUserOfficer(agent)
+        ? sourceProposal.proposerId
+        : agent!.id;
+
+      clonedProposal = await this.proposalDataSource.update({
```
peter-asztalos, line 445: Just a thought: in the future, it may worth refactoring it using transaction, so if any part of the cloning fails we don't end up with a partially cloned proposal.

peter-asztalos (approved):

src/mutations/ProposalMutations.ts
jekabs-karklins: I was actually talking exactly about it during the standup when you were not here :)

PR #257, opened 2021-04-12
Convert organisation and nationality of STFC users

RasmiaKulan:
Closes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/73
### Current Fix
- As part of https://github.com/UserOfficeProject/stfc-user-office-project/issues/73 I have added `1` for `nationality` and `organisation` to add a blank row in the institution table in the database to satisfy the database constraint that the `User` object has.
- Fixed a couple of methods in the  `StfcUserDataSource` that weren't fetching the user details from the web service and hard-coded for `birthdate`, `created`, and `updated` as we don't want to display these user details in the ESS database.
### Future Fix
We need to execute our own STFC specific database patches as we don't need all the data the ESS patches have.


mr-field (changes requested):
I thought we were going to store a dummy record in the Organisation table and fetch details from the UOWS? Looks like this just swaps one hardcoded value for another

RasmiaKulan:
> I thought we were going to store a dummy record in the Organisation table and fetch details from the UOWS? Looks like this just swaps one hardcoded value for another

The dummy row will be a future change when we get our own patches running. For now, it works because the data is filled with the ESS patches. 
The changes I've implemented enables the `me` and `get` function to fetch the user details from the web service which it wasn't doing before.

mr-field:
> The dummy row will be a future change when we get our own patches running. For now, it works because the data is filled with the ESS patches.

Why do we need any patches for this? We can just create a dummy row in the organisation table with the organisation ID from the UOWS. Then, we can fetch the details from the service if necessary. It's the same approach we're using for users - create a dummy record with the user number if one doesn't exist, then use the user number to fetch data from the UOWS.

RasmiaKulan:
As discussed with Panda in the design review, storing the organisation in the database and updating it every time when fetching it from the web service can lead to performance (writing to the database) and conceptual (to a developer) complexity issues. That's why we want to add patches because we need to add a blank row as a one-time operation. 
If we want to store the organisation ID we would have to update it every time we converted the web service object to a user object.

mr-field:
> As discussed with Panda in the design review, storing the organisation in the database and updating it every time when fetching it from the web service can lead to performance (writing to the database) and conceptual (to a developer) complexity issues.

OK, but I'm not talking about storing data in the DB. Only the organisation ID would need to be stored, the rest can be taken from the webservice. The user's organisation can be set to the organisation ID to satisfy the DB contraint, but the data comes from the webservice. In pseudocode:

```
stfcUser = get user from UOWS
stfcOrg = get user organisation from UOWS

essOrg = get DB organisation with id == stfcOrg.id
if essOrg is null {
  create empty organisation with id = stfcOrg.id
}

essUser = get DB user with id == stfcUser.id
if essUser is null {
  create empty user with id = stfcUser.id and organisationId = stfcOrg.id
} else {
  if the organisationId of essUser != stfcOrg.id {
    set essUser.organisationId to stfcOrg.id
  }
}
```

If we run this when the user logs in, they'll be linked to an organisation which has the same ID as their organisation in the UOWS. Then, when we need details on the org, we just fetch them from the service.

RasmiaKulan:
I've made the changes where the organisation ID is fetched from the web service but there are a few things I want to ask:
In the data source, we could get the organisation that was saved when the user first logs in but this wouldn't work if a user has changed their organisation since they first logged in. So in the datasource we then have to check that if it exists and insert it if not - which kind of brings us back to square one. If we are doing that in the datasource, which I think makes sense, is there any point in saving it to the users table when the user logs in? It possibly just adds unnecessary logic in there.
If we don't want to do the checking and inserting of an organisation in the datasource, as Panda mentioned, that is why they were hardcoded as 1 in the user object, so that they'd just reference a blank row. They were changed from 0 to 1 because now that you fixed a couple of the functions to go to the webservice instead of the database, that toEssFunction is actually being used where it should and needs to reference a valid row (0 doesn't exist)

mr-field:
> If we don't want to do the checking and inserting of an organisation in the datasource, as Panda mentioned, that is why they were hardcoded as 1 in the user object, so that they'd just reference a blank row. They were changed from 0 to 1 because now that you fixed a couple of the functions to go to the webservice instead of the database, that toEssFunction is actually being used where it should and needs to reference a valid row (0 doesn't exist)

I'm fairly sure that if all uses are linked to organisation 1, the system will consider all of them to be from the same organisation (i.e. if they go to their details page they'll all see the details for organisation 1). We need to find a way to fetch the right organisation data for all users, and at the moment the system is set to get organisation data by ID from the Organisation table. Since there's a constraint requiring each user to have an organisation, the only way I can think of getting around this is to link the user to the same organisation ID in the ESS DB as the STFC organisation ID they used when they last logged in. We can then use the organisation ID to fetch the correct details from the UOWS.

This can create problems if the user's organisation changes after the last time they logged in, but I don't see any other way to get around the contraint in the short term.

RasmiaKulan:
If we are inserting the organisation in the table can we not do it in the datasource rather than inserting it when a user logs in because the methods in the data source get the user from the web service and convert it to an ess user, we would still have to get the organisation row from the database where the `orgID == webservice.orgID`. We also have to insert it if it doesn't already exist in the datasource, so is there a point in doing this when the user logs in?.
The reason we were leaving the organisation as `1` was that it would reference a blank row once we apply our own patches.

mr-field:
OK, I missed that the user's organisation is already populated from the webservice [here](https://github.com/UserOfficeProject/user-office-backend/blob/cbd5c5a105370389a3e61c9e4c78f66a6cedf657/src/datasources/stfc/StfcUserDataSource.ts#L42) and thought it still had to be loaded using the organisation ID. In this case, always setting the organisation ID to `1` will work for now, as the ID isn't actually used to fetch data. Once we have our own patches, it may be clearer to drop the foreign key constraint entirely rather than adding a blank row, thus removing the need for setting the org ID at all.

However, there is another problem: the `get()` function used to return a user from the ESS DB, and it is used [here](https://github.com/UserOfficeProject/user-office-backend/blob/cbd5c5a105370389a3e61c9e4c78f66a6cedf657/src/mutations/UserMutations.ts#L406) to check if a dummy user exists before creating one. After this PR, the `get()` function will return an STFC user from the UOWS instead, so it will always report that a user already exists. This will prevent the dummy user from being created for new users, who will them be unable to create proposals. To get around this, we could change the [`createDummyUser`](https://github.com/UserOfficeProject/user-office-backend/blob/cbd5c5a105370389a3e61c9e4c78f66a6cedf657/src/datasources/postgres/UserDataSource.ts#L322) function to check if a user already exists before creating one. To then make this clearer, we could rename the function to something like `ensureDummyUserExists`.

jekabs-karklins (approved):

mr-field (approved):

PR #262, opened 2021-04-16
SWAP-1366 If trying to edit a proposal whose call is ended returns an error

peter-asztalos:
## Description

This PR adds some additional checks to prevent users from editing/submitting proposals with an inactive call.

## Motivation and Context

Without this, the user can alter some parts of the proposal and submit it, even though the call is no longer active.

## How Has This Been Tested

- [x] e2e test
- [x] unit test
- [x] manual test 

## Fixes

https://jira.esss.lu.se/browse/SWAP-1366

## Changes

- disallow proposal edit if the call is no longer active
- endpoint added for FE to be able to check if a call is an active call

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/332


simonfernandes (approved):

jekabs-karklins (approved):

jekabs-karklins (changes requested):
I think we should move this logic into QuestionaryAuthorization

src/mutations/QuestionaryMutations.ts
```diff
@@ -75,6 +87,18 @@ export default class QuestionaryMutations {
       return rejection('NOT_FOUND');
     }
 
+    const isUserOfficer = this.userAuth.isUserOfficer(agent);
+    const hasActiveCall = await this.callDataSource.checkActiveCallByQuestionaryId(
+      questionaryId
+    );
+    if (
+      template.templateId === TemplateCategoryId.PROPOSAL_QUESTIONARY &&
```
jekabs-karklins, line 95: This should be like `template.categoryId === TemplateCategoryId.PROPOSAL_QUESTIONARY` , but I think it's best to move these things from Mutation to QuestionaryAuthorizer, it will simplify

martin-trajanovski (approved):

PR #263, opened 2021-04-19
Swap 1374 new edit proposal date field default value resets

jekabs-karklins:
## Description

This PR aims to fix an issue where the default value is lost for date input if other text field changes

## Motivation and Context

This PR simplifies the logic that is behind getting the default value for date input. 
The default value is now returned from the server. 
If the user has not answered the question there will be no answer in the database, if the user has answered the question but left the field empty, the `null` will be saved as the answer. This way we can distinguish between when the user has not answered the question and when he had done it but left the field empty

## Fixes

SWAP-1374

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/334


martin-trajanovski (approved):
Looks good!

PR #264, opened 2021-04-19
71 send emails to busapssink

TCMeldrum:
## Description

https://github.com/UserOfficeProject/stfc-user-office-project/issues/71

Add a few lines to send emails to BISAPPSSINK@stfc.ac.uk if we are in the development environment as set in the environment variable. This is in conjunction with adding an environment variable for duo-backend in docker-orchestration

## Motivation and Context

This is done as a fix to send emails to the sink rather than users. Cosimo has said that we need to do a design review on how we handle emails between us and the ESS. 

## How Has This Been Tested

Emails currently appear in the junk mail of busappssink email. 
Did not seem to break any test as far as I saw.

## Changes

Added conditional statement that checks for NODE_ENV === 'develop' and then if set then sends the email to the sink email with the implementation of the send mail function of \src\eventHandlers\MailService\SMTPMailService.ts

Should be merged with https://github.com/isisbusapps/docker-orchestration/tree/71-adding-sink-email


jekabs-karklins (approved):

simonfernandes (approved):

src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -58,12 +58,18 @@ export class SMTPMailService extends MailService {
             ...(typeof participant.address !== 'string'
               ? {
                   to: {
-                    address: participant.address.email,
+                    address:
+                      process.env.NODE_ENV === 'development'
```
mr-field, line 62: It seems that `NODE_ENV` will be set to `test` rather than `development` when running the tests. I guess this function might be mocked in those cases, but it may be safer to change this to `process.env.NODE_ENV !== 'production'` to make sure we don't send emails while running tests (same thing for the check later).

TCMeldrum: Yeah sounds good, made the change.

mr-field (approved):

PR #270, opened 2021-04-27
53 set authentication in database

AlbertoPaulPopescu:
Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/53?fbclid=IwAR0xrSDaa7DZGVzB0XktksawLKN6SzI-tiUpvioRK9Ls3c9rdihAxfyAcik

The authentication link and status is set in the DB and called at runtime now. The 'Features' table contains the EXTERNAL_AUTH state. The 'Settings' table contains the EXTERNAL_AUTH_LOGIN_URL. The 'process.env.REACT_APP_AUTH_TYPE' and 'process.env.REACT_APP_EXTERNAL_AUTH_LOGIN_URL' was replaced with 'EXTERNAL_AUTH' and 'EXTERNAL_AUTH_LOGIN_URL', respectively, with one exception (App.tsx line 121 - missed).

I had to create a different new pull request for this issue as there were many changes that happened since the first pull request was created and I was receiving multiple errors when working on the old branch: https://github.com/UserOfficeProject/user-office-backend/tree/53-move-auth-variable

1: The settings table was created in the database.
2: 'externalAuth' feature was inserted in the Features table.
3: The code was written following the 'Features' example.
4: Authenticating is fully working 

Known issues: 
1: Had an issue with the 'useDataApi.ts' file, line 59 where I replaced: 'error.response.errors[0].message' with 'error.response.error' to bypass the error and test my changes. However, my fix was reverted back to 'error.response.errors[0].message' before creating the PR.
2: There are still linter errors in the backend but I just followed the already existent convention in the code.
3: In App.tsx I need to replace in line 121 with EXTERNAL_AUTH === true which I missed.

Would like to know:
1: Is the naming convention used fine, any changes/suggestions?
2: Is there any part of the code that could be improved?
3: Is there any old code that needs to be removed as part of this issue?

How Has This Been Tested

1: The new ESS app authentication works as expected when logged in or not logged in into the old P&O system.
2: Tested on Google Chrome.

1: An unit test was added to certify data is retrieved from the database.


db_patches/0091_CreateSettingsTable.sql
```diff
@@ -0,0 +1,23 @@
+DO
+$$
+DECLARE 
+settings_value varchar;
```
martin-trajanovski, line 4: Is this declaration needed here on the top? I cannot see that it is used after its declaration. If it is not needed you can remove the whole declaration from the top.

src/queries/AdminQueries.ts
```diff
@@ -41,6 +41,10 @@ export default class AdminQueries {
     return await this.dataSource.getFeatures();
   }
 
+  async getSettings() {
```
martin-trajanovski, line 44: I guess this should be open endpoint for anyone to access. There is no authorization needed.

martin-trajanovski (approved):
Otherwise it looks great! 👀 💯

db_patches/0091_CreateSettingsTable.sql
AlbertoPaulPopescu: @martin-trajanovski  Thank you, I will remove it :)

src/queries/AdminQueries.ts
AlbertoPaulPopescu: @martin-trajanovski Is it necessary to create a separate queries.ts file or is there an existing one where I can place my query in?

martin-trajanovski: Not sure how do you think create separate queries file. My comment was just to confirm that this should be an open endpoint. If that is the case you don't need to change anything. Sorry if it was a bit confusing 😄

AlbertoPaulPopescu: @martin-trajanovski I misinterpreted what you meant by that, I thought you wanted me to move the query to a separate file 😁

db_patches/0092_SetAuthFeature.sql
```diff
@@ -0,0 +1,14 @@
+DO
+$$
+BEGIN
+	IF register_patch('SetAuthFeature.sql', 'Will Edwards', 'Set STFC auth to enabled', '2021-02-23') THEN
```
JEThompson97, line 4: Is the **name and date** on this correct?
I would have thought it should be your name?
(wanted to check in case this was copied from somewhere and not fully updated)

JEThompson97 (approved):
Otherwise looks good 🙂

db_patches/0092_SetAuthFeature.sql
AlbertoPaulPopescu: This was the part of the issue that Will has done, so I kept his name and date.

JEThompson97: Ahh I see, thanks for clarifying 🙂

PR #278, opened 2021-05-04
As a beamline manager I should get all technical reviews and be able to reassign them to other instrument scientist

jekabs-karklins:
## Description

This PR aims to fix the user story where it is necessary to have a responsible person for the technical review.
image

## Motivation and Context

- This PR adds a new property for the instrument called "beamline manager".  
- The responsible person for the technical review is initially always set as a beamline manager.
- It is possible for a beamline manager, user officer, or responsible user to re-assign the task to someone else.
- The information about who did the review is visible to the SEP panel.

## How Has This Been Tested

Added E2E

## Fixes

SWAP-1518

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/360


src/mutations/ReviewMutations.ts
```diff
@@ -238,6 +249,37 @@ export default class ReviewMutations {
       });
   }
 
+  async isTechnicalReviewAssignee(
```
martin-trajanovski, line 252: I think this function should be part of the `UserAuthorization` utils. For me it sounds more like if somebody has some rights to access something and that is the usual place where we keep the things like that.

martin-trajanovski (approved):
Looks great. Just a small change maybe required.

src/mutations/InstrumentMutations.ts
```diff
@@ -139,14 +142,31 @@ export default class InstrumentMutations {
     );
 
     if (!allProposalsAreOnSameCallAsInstrument) {
+      logger.logError(
+        'Can not assign the proposal to the instrument because proposals call has no such intrument',
```
JEThompson97, line 146: 'instrument' typo, plus optional tweaks
```suggestion
        'Cannot assign the proposal to the instrument because the proposals call has no such instrument',
```

JEThompson97 (approved):
Code looks good as far as I can tell 🙂

PR #281, opened 2021-05-06
UserOfficeProject/stfc-user-office-project#109: Update check for if t…

Ziggiyzoo:
…he proposal is submitted

## Description
Closes UserOfficeProject/stfc-user-office-project#109

This allows the proposal to be updated after it has been submitted if the status is "EDITABLE_SUBMITTED". Updates the `ProposalMutations.ts` file to ignore the check for the proposals submitted boolean if its status is "EDITABLE_SUBMITTED".

## Motivation and Context

The editable submitted proposal was not saving after being edited.  This fixes this and allows the proposal to save.

## How Has This Been Tested

I have tested:
- Changing proposal name
- Changing proposal text
- Changing proposal co proposers

Some issues I noticed while testing:

- All proposals (using the default workflow) will continue to have the draft status after being submitted. They are editable, but cannot be saved. Two ways to fix this, edit the logic behind making the proposal read-only, OR,  Add in a status of "SUBMITTED" and have the default workflow assign this after submission.


src/mutations/ProposalMutations.ts
```diff
@@ -124,9 +124,10 @@ export default class ProposalMutations {
     ) {
       return rejection('NOT_ALLOWED');
     }
-
-    if (proposal.submitted && !this.userAuth.isUserOfficer(agent)) {
-      return rejection('NOT_ALLOWED_PROPOSAL_SUBMITTED');
+    if (!(proposal.statusId.valueOf() == 14)) {
```
mr-field, line 127: Is there a way to do this check using the `EDITABLE_SUBMITTED` short code, instead of the status ID? It would be clearer to read and avoid depending on a fixed ID that may change

martin-trajanovski: For now it is fine to leave it like this. As I mentioned this requires a bit more changes to make it right. Maybe you can get the status by `proposal.statusId` and then compare the returned one by status shortCode. But to achieve this you might need to inject the `ProposalSettingsDataSource` into the `ProposalMutations` itself and then you can query `getProposalStatus` for example to get the status with shortCode.

martin-trajanovski (approved):

jekabs-karklins (approved):

mr-field (approved):

mr-field:
I think you'll need to merge `develop` into this again and solve the merge conflict before the tests can run, but the PR looks good to me.

Ziggiyzoo:
It looks like the tests expect `Can not update proposal after submission`, but there is no value for this is the file. Instead `NOT_ALLOWED_PROPOSAL_SUBMITTED` is returned... and this is used by the frontend, so I am not sure if this was changed and somehow skipped the tests when being added?

PR #284, opened 2021-05-12
try new tag and release action

martin-trajanovski:
## Description

A GitHub Action to automatically bump and tag master, on merge of a pull request.

## Motivation and Context

Previously this was done manually to tag and create a release.

## How Has This Been Tested

Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1474

## Changes

Added new github action called `github-tag-and-release`

### DIFF
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -41,7 +41,9 @@ jobs:
       - name: Extracted branch
         run: echo "Extracted branch ${{ steps.extract_branch.outputs.branch }}"
 
+      # Build and push only for 'develop' branch because for 'master' we have another action that builds release image
       - name: Build and push
+        if: steps.extract_branch.outputs.branch == 'develop'
         id: docker_build
         uses: docker/build-push-action@v2
         with:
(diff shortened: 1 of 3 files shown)


jekabs-karklins (commented):
I see that in order for script to decide on correct semantic versioning number for the things merged in master the commit messages needs to be formatted accordingly (https://github.com/mathieudutour/github-tag-action). 

Maybe we can make sure that when merging into master there  must be always a merge commit ( if that is possible ) and put the rules for how to format  the message in the readme file? Otherwise the versioning will not be following conventions.

martin-trajanovski:
> I see that in order for script to decide on correct semantic versioning number for the things merged in master the commit messages needs to be formatted accordingly (https://github.com/mathieudutour/github-tag-action).
> 
> Maybe we can make sure that when merging into master there must be always a merge commit ( if that is possible ) and put the rules for how to format the message in the readme file? Otherwise the versioning will not be following conventions.

Yes you are right! I wanted to mention this but forgot. Here are some guidelines that we should follow (https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines) for our commits so the generation of version number will be correct depending on that. This should be written somewhere as documentation so everybody is aware of the right commit messages to use.

bolmsten (approved):

martin-trajanovski:
> I see that in order for script to decide on correct semantic versioning number for the things merged in master the commit messages needs to be formatted accordingly (https://github.com/mathieudutour/github-tag-action).
> 
> Maybe we can make sure that when merging into master there must be always a merge commit ( if that is possible ) and put the rules for how to format the message in the readme file? Otherwise the versioning will not be following conventions.

I created a task for it in our Jira backlog so we wont forget it. https://jira.esss.lu.se/browse/SWAP-1579

PR #289, opened 2021-05-18
Swap 1474 tag and release master

martin-trajanovski:
## Description

Merge the changes for tagging and releasing into develop as well.


.github/workflows/build-push.yml
```diff
@@ -41,7 +41,9 @@ jobs:
       - name: Extracted branch
         run: echo "Extracted branch ${{ steps.extract_branch.outputs.branch }}"
 
+      # Build and push only for 'develop' branch because for 'master' we have another action that builds release image
       - name: Build and push
+        if: steps.extract_branch.outputs.branch == 'develop'
```
jekabs-karklins, line 46: Don't we want to push also other branches than `develop`. In order for E2E tests to work frontend will pull image tagged with the branch name same as frontend branch, or is this handled somewhere else?

martin-trajanovski: Yes you are right. Good point! I should update this maybe to be just something like `!= master`

martin-trajanovski: @jekabs-karklins I guess now it is much better and should be fixed.

jekabs-karklins (approved):

jekabs-karklins (approved):

PR #290, opened 2021-05-20
feat(sample shipments): Add proposal shortCode to sample shipments

martin-trajanovski:
## Description

Added new resolver in the shipment just to return proposal `shortCode` and show it in the UI.

## How Has This Been Tested

e2e tests
manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1477


src/datasources/postgres/ShipmentDataSource.ts
```diff
@@ -91,7 +91,7 @@ export default class PostgresShipmentDataSource implements ShipmentDataSource {
       });
   }
 
-  update(args: UpdateShipmentArgs): Promise<Shipment> {
+  async update(args: UpdateShipmentArgs): Promise<Shipment> {
```
jekabs-karklins, line 94: Are these added for consistency?

jekabs-karklins (approved):
Looks good :)

src/datasources/postgres/ShipmentDataSource.ts
martin-trajanovski: Yes because there is no difference anyway but it looks better when all functions are with the async prefix because they are actually asyc anyway

PR #291, opened 2021-05-20
Move sensitive data into an external config

RasmiaKulan:
Closes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/55
### Changes made
1. Deleted `config.json` file (as discussed)
2. Added an `env-loader.js` file that loads the path to the external configuration and checks the environment the application is running on (dev or prod).
3. Amended the `index.ts` file to import the env-loader. This was done so the config script is run before referencing any environment variables.
4. Renamed `example.env` to `example.development.env` because the `env-loader.js` checks an environment to run the .env file for eg. `development.env` or `production.env`.

### Actions required
1. In the `docker-compose` file set the `CONFIG_PATH` to the destination where the sensitive data is located for eg. `C:/path/to/config`
3. Create separate configs for development `(development.env)` and production `(production.env)`


JEThompson97 (approved):
Looks good I think

mr-field (approved):

martin-trajanovski (approved):

PR #292, opened 2021-05-21
feat: Swap 1570 as a user i want to be able to register a visit

jekabs-karklins:
## Description

As a user I want to be able to register a visit

## Motivation and Context

When user has a proposal accepted, he should be able to register for a visit. 
The registration form is based on templating system, that user officer can design.

## Fixes

SWAP-1517

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/377


src/datasources/postgres/VisitationDataSource.ts
```diff
@@ -0,0 +1,115 @@
...
+import database from './database';
+import {
+  createBasicUserObject,
+  createVisitationObject,
+  VisitationRecord,
+} from './records';
+
+class PostgresVisitationDataSource implements VisitationDataSource {
+  getTeam(visitationId: number): Promise<BasicUserDetails[]> {
+    console.log(
```
martin-trajanovski, line 15: You have some console.log forgotten here.

src/mutations/VisitationMutations.ts
```diff
@@ -0,0 +1,133 @@
...
+        'Can not update visitation because of insufficient permissions',
+        { user, visitationId }
+      );
+    }
+
+    return this.dataSource.deleteVisitation(visitationId);
+  }
+
+  async hasWriteRights(user: UserWithRole | null, visitationId: number) {
+    return this.visitationAuthorization.hasWriteRights(user, visitationId);
```
martin-trajanovski, line 131: I think there is no need for creating new function called `hasWriteRights` when we already have that so we can just use `this.visitationAuthorization.hasWriteRights` in the places where we need it is one-liner anyway. Like this we create 3 more lines of code but it is not simplifying the things. It is again up to you but I think there is no need to duplicate one function when we already have function that can be reused.

martin-trajanovski (approved):
In general it looks good 👍

src/datasources/postgres/VisitationDataSource.ts
jekabs-karklins: Check! :) Thanks

src/mutations/VisitationMutations.ts
jekabs-karklins: Having function actually is little more readable in my oppinion in this case because of the line length limitation we have. It is easier to read if invoking shorter function `hasWriteRights` then  `this.visitationAuthorization.hasWriteRights` so it does not wrap into multiple lines.

PR #303, opened 2021-05-25
120 win images built in gh actions

TCMeldrum:
## Description

Added push-build-win task for github actions. The official docker-build-push is linux only so have used a 3rd party one by mr-smithers-excellent which works and doesn't seem to anything shady. Currently hardcoded to run on windows-2019 runners, not windows-latest to avoid versioning issues if windows-latest changes. 

I have updated our dockerfiles to point at dockerhub rather than the private repository. 

Sorry about the commits in the backend  I tried to use rebased to remove some of my debugging around the github action but I might have messed up somewhere if you see anything that has changed/reverted that shouldn't have please tell me.

## Motivation and Context

By pushing windows image to dockerhub we can now run tests on windows on github actions.

## How Has This Been Tested

Images push to dockerhub correctly and new docker files should be functionally the same.

## Depends on

Should probably go in with the PR's for other build tasks all are basically the same as here
https://github.com/UserOfficeProject/user-office-frontend/pull/384
https://github.com/UserOfficeProject/user-office-scheduler-backend/pull/34
https://github.com/UserOfficeProject/user-office-gateway/pull/18
https://github.com/UserOfficeProject/user-office-factory/pull/26

Closes issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/120


martin-trajanovski (approved):

.github/workflows/build-push-win.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Build and push
+        id: docker_build
+        uses: mr-smithers-excellent/docker-build-push@v5
+        with:
+          username: ${{ secrets.DOCKER_STFC_USERNAME }}
+          password: ${{ secrets.DOCKER_STFC_PASSWORD }}
+          registry: docker.io
+          buildArgs: BUILD_VERSION=${{ github.sha }}, WINDOWS_VERSION=ltsc2019
+          dockerfile: ./Dockerfile.win
+          image: dmsc/duo-backend-win
```
mr-field, line 44: The Linux action seems to tag the built image with the branch name (see [here](https://github.com/UserOfficeProject/user-office-backend/blob/f7b6970e6b5704b15bb65d159f72cc13cd1677db/.github/workflows/build-push.yml#L50)), which avoids overwriting the default image on every pull request. I think we want to do the same, otherwise every change for review will be pushed as the stable image. Can we reuse `{{ steps.extract_branch.outputs.branch }}` for that?

TCMeldrum: Sorry, I meant for that to be the case but must have missed it when I change the push build action I'll change that now. 

Looking at the documentation of the push build action it seems that it does some automatic tagging with the branch name so they are tagged on dockerhub with the branch name and reference number, so I thought it was working as intended.

mr-field: It's true that images would be tagged with the branch name and commit hash automatically, but I think we want to mirror the Linux behaviour and only use the branch name. I think there's some tests which expect the backend and frontend images to have the same tag, which they wouldn't have if the commit hash was included.

mr-field (approved):

PR #307, opened 2021-05-27
Swap 1433 sep proposal assignment UI improvement

martin-trajanovski:
## Description

Improve assigning and removing SEP to proposals.

## How Has This Been Tested

e2e tests
manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1433


src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -233,17 +255,13 @@ export default class PostgresInstrumentDataSource
...
-      `Could not assign proposals ${proposalIds} to instrument with id: ${instrumentId} `
+      `Could not assign proposals ${proposalIds} to instrument with id: ${instrumentId}`
     );
   }
 
-  async removeProposalFromInstrument(
-    proposalId: number,
-    instrumentId: number
-  ): Promise<boolean> {
+  async removeProposalsFromInstrument(proposalIds: number[]): Promise<boolean> {
```
jekabs-karklins, line 262: Don't you need an instrument ID any more? Can proposal be assigned only to one instrument?

jekabs-karklins (approved):

src/datasources/postgres/InstrumentDataSource.ts
martin-trajanovski: Yes currently it can be assigned to only one instrument that's why I removed the instrumentId. If we need something like one proposal to be assigned to more instruments later then we will need more changes.

src/resolvers/mutations/AssignProposalsToSep.ts
```diff
@@ -14,36 +14,46 @@ import {
   SEPResponseWrap,
 } from '../types/CommonWrappers';
 import { wrapResponse } from '../wrapResponse';
+import { ProposalIdWithCallId } from './ChangeProposalsStatusMutation';
 
 @ArgsType()
-export class AssignProposalToSEPArgs {
+export class AssignProposalsToSepArgs {
+  @Field(() => [ProposalIdWithCallId])
+  public proposals: ProposalIdWithCallId[];
```
jekabs-karklins, line 22: Why are we sending in tuple with ProposalId and CallId. Isn't a proposal already attached to a call?
I believe only thing we need here is a sepId and an array of proposalIds.

martin-trajanovski: It is much easier like this I think. Because if I send just proposal ids then I need to get call ids by querying the proposals by id and get call id for each proposal which complicates the things a bit. Like this we already have them on the frontend when we select and we can easily send them.

PR #316, opened 2021-06-01
feat(calls and proposals): Add proposal allocation time unit on calls

martin-trajanovski:
## Description

Added allocation time unit on the calls and possibility to set it.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1431


db_patches/0096_AddAllocationTimeUnitToCallTable.sql
```diff
@@ -0,0 +1,13 @@
+DO
+$$
+BEGIN
+	IF register_patch('AddAllocationTimeUnitToCallTable.sql', 'martintrajanovski', 'Add proposal allocation time unit to call table', '2021-05-31') THEN
+		BEGIN
+
+			ALTER TABLE call ADD COLUMN allocation_time_unit INTEGER NOT NULL DEFAULT 0;
```
jekabs-karklins, line 7: I think it would be better to set 'day' or 'hour' than number. Also add a check to it

ALTER TABLE call ADD COLUMN allocation_time_unit VARCHAR(10) CHECK(allocation_time_unit IN ('day', 'year')) NOT NULL DEFAULT 'day';

martin-trajanovski: Not really sure why the `day` or `hour` is better than 0 or 1. Is it just because it is more descriptive or?

martin-trajanovski:
@jekabs-karklins this is changed as you requested using `day` and `hour` instead of 0 and 1.

jekabs-karklins (approved):

PR #320, opened 2021-06-03
Move env-loader in order to fix STFC dev backend deployment

simonfernandes:
Closes UserOfficeProject/stfc-user-office-project#138

## Description

This moves the env loader file into the src directory. It (hopefully) fixes our dev deployment because the file wasn't being copied in our dev dockerfile. It's been fixed like this (without modifying the dockerfile) as it's probably a more logical place for it to be.

## How Has This Been Tested

I've tested that this works when loading a config from the CONFIG_PATH, and also from a .env file in the root. So no changes to how they are handled is necessary.


mr-field (commented):
Are we definitely sure that [this import](https://github.com/UserOfficeProject/user-office-backend/blob/42bc4c352d54eb73ab307a1b7f6754b6dd802923/index.ts#L5) doesn't need updating? Even if it works without changing it, maybe it's worth doing it just for clarity.

simonfernandes:
@mr-field Sorry, it absolutely does need updating and I'd forgotten to include it in the commit. Thanks for noticing

mr-field (approved):

martin-trajanovski (approved):

PR #95, opened 2020-11-30
SWAP-982 #resolve and fix some call validation problems

martin-trajanovski:
## Description

Create/Update call validation schemas updated and split to fit the frontend needs. Here we just merge them and apply everything at once

## Motivation and Context

The point is to have step validation on the frontend and full validation on the backend.

## How Has This Been Tested

With e2e tests on the frontend.

## Fixes

https://jira.esss.lu.se/browse/SWAP-982

## Depends on

https://github.com/UserOfficeProject/user-office-lib/pull/18

## Tests included/Docs Updated?

- [x] I have added tests to cover my changes.
- [ ] All relevant doc has been updated

PR #326, opened 2021-06-08
95 different features for stfc and ess

TCMeldrum:
## Description

Ref: https://github.com/UserOfficeProject/stfc-user-office-project/issues/95

Added featureConfig.ts to src, this reads in featureConfig.json and then disables/enables features in the feature database. Added example config file for STFC and ESS, the default config should pass the automated tests. 

This solution doesn't separate any of the db patches which means we might be adding data to the database just to change it. If anyone has a better solution please don't hesitate to share.

This may be blocked by this issue:  https://github.com/UserOfficeProject/stfc-user-office-project/issues/54 as it will be expanding the features table. 

## Motivation and Context

This is to allow facility-specific feature enabling through a simple config file.
Issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/95

## How Has This Been Tested

Config file changes things within the features table, both example config files work however the number of features within the table is small.

## Changes

Introduced featureConfig.ts and added the config files as well as example config files for ESS and SFTC. This featureConfig.ts is imported in the index file for the backend. 

Issue #95 also mentions data that we might be facility-specific although we could not find anything relevant at this time


RasmiaKulan (approved):

src/featureConfig.ts
```diff
@@ -0,0 +1,18 @@
+import * as config from '../featureConfig.json';
+import database from './datasources/postgres/database';
+
+function setFeatures() {
```
simonfernandes, line 4: It looks good! Does `index.ts` just run everything in this file when it's imported?

It's not a big deal, but I wonder if a better place to have this function might be in the `AdminDataSource`? So it could `applyPatches()` and set features around the same time. Or is that too late in the process to toggle features?

TCMeldrum: Yeah, runs a file when it is imported. 

I was struggling to work out where to run this code. I chose here as env-loader.js is performing a similar function in setting the environment so thought here would make sense but `AdminDataSource` would make sense as well. I don't think it would be too late in the process so  I don't mind moving it.

TCMeldrum:
@UserOfficeProject/swap would one of you be able to look over this as well?

jekabs-karklins (approved):
Hi, we talked internally a little about this PR and I believe this functinality could be simplified if we'd use the dependecy injection, as DI has already the needed means for separating functionality between facilites. I have made a quick PR here https://github.com/UserOfficeProject/user-office-backend/pull/336/files which seems to work without any issues.

I have a comment on this PR. I can see that featureConfig.json is added to .dockerignore, wouldn't that mean that there would be some issues during the runtime as the file would be excluded from the docker context?

jekabs-karklins (commented):
Hi, we talked internally a little about this PR and I believe this functinality could be simplified if we'd use the dependecy injection, as DI has already the needed means for separating functionality between facilites. I have made a quick PR here https://github.com/UserOfficeProject/user-office-backend/pull/336/files which seems to work without any issues.

I have a comment on this PR. I can see that featureConfig.json is added to .dockerignore, wouldn't that mean that there would be some issues during the runtime as the file would be excluded from the docker context?

TCMeldrum:
Had a look through the PR and dependency injection seems like a good simpler solution. I would be happy to use it instead. I liked the idea of the external file and it would be easy to see which features were disabled/enabled as the features table grows with this issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/54 but hard coding them work as well.

For the .dockerignore it is telling docker to not ignore the file with the ! as in .dockerignore I think we tell it to ignore everything except what we specify.

TCMeldrum:
Used https://github.com/UserOfficeProject/user-office-backend/pull/336 instead

PR #333, opened 2021-06-14
fix: bugfix for inviting users and improve some logic

martin-trajanovski:
## Description

Fixing bug with user invitation and improving some logic when user is invited with `SEP_reviewer` role. Also making all role short codes consistent because some were uppercase some lowercase.

## How Has This Been Tested

manual tests
e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1612


jekabs-karklins (approved):
💯

PR #336, opened 2021-06-15
patch: Differentiate features #suggestion

jekabs-karklins:
## Description

This PR aims to fix the issue of different facilities needing different features enabled/disabled by default.

## Motivation and Context

It uses dependency injections where different functions implementing features enabling are run depending on the facility.

## Depends on

N/A

### DIFF
--- a/db_patches/0076_CreateFeatureToggle.sql
+++ b/db_patches/0076_CreateFeatureToggle.sql
@@ -13,9 +13,9 @@ BEGIN
       ); 
 
       INSERT INTO 
-        features(feature_id, is_enabled, description)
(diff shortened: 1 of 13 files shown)


TCMeldrum:
Copying part of my comment from the other PR: 

Had a look through the PR and dependency injection seems like a good simpler solution. I would be happy to use it instead. I liked the idea of the external file and it would be easy to see which features were disabled/enabled as the features table grows with this issue: UserOfficeProject/stfc-user-office-project#54 but hard-coding them work as well.

simonfernandes (approved):
Looks very clean

PR #342, opened 2021-06-25
81 update proposal editor interface to improve clarity

simonfernandes:
## Description

Frontend branch: https://github.com/UserOfficeProject/user-office-frontend/pull/403

## Motivation and Context

To enable checking whether a proposal will have an upcoming editable submitted status and display a different submit button confirmation to the user.

## How Has This Been Tested

Tested manually

## Changes

- Remove user auth to enable the editable submitted check from a user account
- Made status changing events on a connection optional, as not all connections have them (the default draft status for example)

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/403


src/queries/ProposalSettingsQueries.ts
```diff
@@ -83,7 +83,6 @@ export default class ProposalSettingsQueries {
     return groupedProposalWorkflowConnections;
   }
 
-  @Authorized([Roles.USER_OFFICER])
```
jekabs-karklins, line 86: @martin-trajanovski  what do you think about this? Is it ok to remove it?

martin-trajanovski: Hmm not sure why it needs to be removed. Because for now we allow only user officers to create workflow groups. Not sure what is the initial reason for removal.

jekabs-karklins: It says in the description:
`Remove user auth to enable the editable submitted check from a user account`
I was wondering @martin-trajanovski if there are any security risks for exposing this?

martin-trajanovski: I think it should keep at least the `Authorized` without any role just to prevent any risks from unauthorized access to some data.

martin-trajanovski (changes requested):
Keep `Authorized` decorator at the `proposalWorkflowConnectionGroups` query just remove the role to prevent unauthorized access to the data.

martin-trajanovski (approved):

PR #349, opened 2021-07-09
Add STFC theme

simonfernandes:
## Description

Backend branch for https://github.com/UserOfficeProject/user-office-frontend/pull/412

This adds the theme variables to the settings table so that they can be read and applied.

>This refactors the theme so that it now reads colours from variables in the Settings table of the database. If there are colours defined in any of the theme variables, the Material UI theme is updated as well as the CSS variables. If no colours are defined in the database, the default Material UI colours are used.

>This also allows for an optional header image. The filename can be specified in the Settings table and it will be used as an import by looking for the file in the images directory.

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/412


simonfernandes:
Closed because I accidentally named the branch incorrectly so it wasn't picked up by the frontend

PR #352, opened 2021-07-12
fix(tepmlates): Improving archived templates queries

martin-trajanovski:
## Description

Improved `getTemplates` query so that it returns archived template only if `templateIds` is passed to show it in call/other template where it is already used but after that it is archived. Previously if template gets archived and it is already used in a call or another template empty dropdown was shown because it wasn't returned from the query. We were returning only unarchived templates.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1412


src/datasources/postgres/TemplateDataSource.ts
```diff
@@ -151,7 +153,7 @@ export default class PostgresTemplateDataSource implements TemplateDataSource {
       });
   }
 
-  async getTemplate(templateId: number) {
+  async getTemplate(templateId?: number) {
```
jekabs-karklins, line 156: I think `templateId` should not be optional for the getTemplate

It it is not set, it will return unexpected results

src/queries/TemplateQueries.ts
```diff
@@ -12,7 +12,7 @@ import { TemplatesArgs } from '../resolvers/queries/TemplatesQuery';
 @injectable()
 export default class TemplateQueries {
   constructor(
-    @inject(Tokens.TemplateDataSource) private dataSource: TemplateDataSource
+    @inject(Tokens.TemplateDataSource) public dataSource: TemplateDataSource
```
jekabs-karklins, line 15: I think it is better to keep the parameter private and resolve the dependency where you need it with
`container.resolve(TemplateDataSource)`

src/resolvers/queries/ProposalTemplatesQuery.ts
```diff
@@ -15,6 +16,9 @@ import { ProposalTemplate } from '../types/ProposalTemplate';
 class ProposalTemplatesFilter {
   @Field({ nullable: true })
   public isArchived?: boolean;
+
+  @Field(() => [Int], { nullable: true })
```
jekabs-karklins, line 20: This should be `@Field(() => [Int!], { nullable: true })`

jekabs-karklins (changes requested):

src/resolvers/queries/ProposalTemplatesQuery.ts
martin-trajanovski: I think this is how it should be on the frontend here it is not allowed to use `Int!`

src/datasources/postgres/TemplateDataSource.ts
```diff
@@ -133,13 +133,15 @@ export default class PostgresTemplateDataSource implements TemplateDataSource {
   async getTemplates(args: TemplatesArgs): Promise<Template[]> {
     return database('templates')
       .select('*')
-      .where({ is_archived: args.filter?.isArchived || false })
       .modify((query) => {
+        if (args.filter?.isArchived !== undefined) {
+          query.where({ is_archived: args.filter?.isArchived });
```
jekabs-karklins, line 138: I think you can skip optional chaining since it's already checked in the if statement
`query.where({ is_archived: args.filter.isArchived });`

jekabs-karklins (approved):

src/resolvers/types/Call.ts
```diff
@@ -71,8 +74,8 @@ export class Call implements Partial<CallOrigin> {
   @Field(() => AllocationTimeUnits)
   public allocationTimeUnit: AllocationTimeUnits;
 
-  @Field(() => Int, { nullable: true })
-  public templateId?: number;
+  @Field(() => Int)
```
jekabs-karklins, line 77: If this is not optional, then I suggest we also mark it as non-optional in the `createCallMutation`

martin-trajanovski: It is done in the next PR. It is a bit separated but the next one is more about that making template as reqiured. This one is just for showing archived templates.

PR #358, opened 2021-07-16
Swap 1696 - Broadcast messages for proposal events such as creation and update

bolmsten:
## Description

There are a number of services interested in proposal events, with this PR they can listen to the rabbitmq broadcasting channel for events like proposal_update, proposal_created etc. 

## Motivation and Context

Integration with DMP, SciCat and SciChat to start with

## Fixes

SWAP-1696


src/mutations/QuestionaryMutations.ts
```diff
@@ -54,6 +55,7 @@ export default class QuestionaryMutations {
   }
 
   @Authorized()
+  @EventBus(Event.PROPOSAL_TOPIC_ANSWERED)
```
jekabs-karklins, line 58: I remember we talked about if the event should be called PROPOSAL_TOPIC_ANSWERED or TOPIC_ANSWERED.

jekabs-karklins (approved):
LGTM

PR #360, opened 2021-07-21
Swap 1277 use same validation for questionary components

martin-trajanovski:
## Description

Moved validation for questionary components in the duo-validation package to reuse on both backend and frontend.

## Motivation and Context

Previously we had validations separated on the frontend and backend and sometimes they are out of sync.

## How Has This Been Tested

e2e tests
manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1277


src/models/questionTypes/RichTextInput.ts
```diff
@@ -51,55 +52,14 @@ export const sanitizerConfig: IOptions = {
...
     }
-    const config = field.config as RichTextInputConfig;
-    if (config.required && !value) {
-      return false;
-    }
-
-    if (
-      config.max &&
-      value &&
-      sanitizeAndCleanHtmlTags(value).length > config.max
```
jekabs-karklins, line 97: I'm thinking if we still need to do this? I believe Peter had put it in for a reason

jekabs-karklins (approved):
💯

src/models/questionTypes/RichTextInput.ts
martin-trajanovski: There is `sanitizeHtml(value, sanitizerConfig)` in the transform function that Peter has created to cleanup the input but this for validation I added this sanitizeAndCleanHtmlTags which now it is transfered to the validation package as it was here and we use the same one for both frontend and backend.

PR #365, opened 2021-07-26
SWAP-1681-use-call-allocation-unit-to-calculate-allocated-time

martin-trajanovski:
## Description

Use allocation unit to calculate the right allocation time for scheduler

## Motivation and Context

Previously the allocated time in the scheduler was calculated only in days now if hours is selected as allocation time unit on the call the allocated time is calculated properly.

## How Has This Been Tested

- manual tests
- currently we can't access the scheduler from the core and it is not possible to cover this by e2e tests but later when we fix this will be good to write some e2e tests.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1681


src/eventHandlers/messageBroker.ts
```diff
@@ -126,12 +131,29 @@ export function createPostToRabbitMQHandler() {
...
+        if (!call) {
+          logger.logWarn(`Proposal '${proposal.primaryKey}' has no call`, {
+            proposal,
+          });
+
+          return;
+        }
+
+        // NOTE: Default allocation unit is Days. The UI supports Days and Hours.
+        let proposalAllocatedTime =
```
jekabs-karklins, line 145: Maybe it's more readable if you added a method `getSecondsPerAllocationTimeUnit(unit:AllocationTimeUnits) :number`

Just a suggestion.

jekabs-karklins (approved):

PR #368, opened 2021-07-29
Add proposal details to log statement

vyshnavi1605:
## Description

I added 2 log statements which log the `proposal id`, `proposal title` and `proposer id` (reference number) when the user submits or updates a proposal. This will then be visible to us on the log server to search up. 

[Wiki Page - linking to this issue](https://github.com/isisbusapps/ISISBusApps/wiki/Proposal-submission-disaster-recovery)
The link above to the wiki explains, how we can retrieve the data on what users have submitted and updated. 

## Motivation and Context

This change is required as, in case of a tragic event where we loose data from the database which has not been backed up, we can check what users have lost data through the log server.  

## How Has This Been Tested

To test my implementation, I have submitted and updated a proposal on the new proposal system and then checked in the backend container if the log statement I added has executed.  
image
The image above shows the backend container output when a user updates a proposal 

## Fixes

Fixes https://github.com/UserOfficeProject/stfc-user-office-project/issues/162

## Changes

I added a log statement in the ProposalsMutations file in the backend folder. I added the statements in the updated and submitted proposals return. 


src/mutations/ProposalMutations.ts
```diff
@@ -156,13 +157,30 @@ export default class ProposalMutations {
...
     }
 
-    return this.proposalDataSource.update(proposal).catch((err) => {
-      return rejection(
-        'Could not update proposal',
-        { agent, primaryKey: proposalPk },
-        err
-      );
-    });
+    return this.proposalDataSource
```
TickleThePanda, line 160: Could these be done as an [`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await)? I think it would read more cleanly. Something like this below should work:
```ts
try {
  const updatedProposal = await this.proposalDataSource.update(proposal);
  logger.logInfo(...);
  return updatedProposal;
} catch (err) {
  throw new Error(...);
}
```

TickleThePanda (changes requested):

martin-trajanovski (approved):

TickleThePanda (approved):

PR #385, opened 2021-08-12
Added attachment for footer logo

vyshnavi1605:
## Description

Closes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/52
I added an image attachment into the email template. So that there will be an image displayed in the footer of the confirmation email. 

## How Has This Been Tested
The way this has been done is through submitting proposals on the new proposals system. Once a proposal is submitted the BusAppsSinks inbox receives a confirmation email that I checked to see if it had the correct format. 

## Changes
- Added the links to the email and changed the wording of the email in the `html.pug` file
- Added an image to the footer which was then added to `html.pug`.
- Added the `styles.css` sheet and sized the logo. 

Branch Name: stfc-user-office-project_52_amend_confirmation_email_format

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,8 @@
+FROM node:12-alpine
+
+# The Git repo should be mounted from the host into this directory
+WORKDIR /app
+
+EXPOSE 4000
+
+CMD ["npm", "run", "dev:docker"]


TickleThePanda (changes requested):

TCMeldrum:
What should the logo path be on dev? EMAIL_FOOTER_IMAGE_PATH isn't set in the dev config file. Do we also need to add these changes to BIS\kubernetes-shared-dev\BisAppSettings or is that done automatically as part of the dev deployment?

vyshnavi1605:
> What should the logo path be on dev? EMAIL_FOOTER_IMAGE_PATH isn't set in the dev config file. Do we also need to add these changes to BIS\kubernetes-shared-dev\BisAppSettings or is that done automatically as part of the dev deployment?

I added the logo path only in the local dev file. I'm not entirely sure if it will get added automatically, I can add the path to both or just the dev config file. What would you suggest?

TickleThePanda (approved):

TickleThePanda:
I don't think the ESS team need to look at this one so you're welcome to merge whenever you're ready.

TCMeldrum (approved):

PR #398, opened 2021-09-03
Added prod service and prod feature enabling

TCMeldrum:
Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/192

## Description

This is my suggestion of how STFC can automatically update features/settings on production while leaving the ESS features on production unchanged. Using Tsyringe to create ConfigureProductionEnvironment function that is called from within the index file. For all dependency configs apart from STFC, this is an empty function.

This might be overcomplicating things a little but I felt it allow for more flexible deployments between production and development in the future.

Also added some stfc specific configuration for production. 

## Motivation and Context

Allowing STFC to set features in production automatically.

## Fixes

Issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/192

## Depends on

Docker-orchestration: https://github.com/isisbusapps/docker-orchestration/pull/150


mr-field (approved):
The changes for automatically setting the features look OK to me. It may be best to post a link to this PR in the Slack thread about settings, so that someone from the ESS side can also have a look.

More strictly related to https://github.com/UserOfficeProject/stfc-user-office-project/issues/192, we may have an additional problem: the UOWS client is auto-generated from the service WSDL using [this](https://github.com/UserOfficeProject/user-office-backend/blob/6145dadbc568395a0c84bd6042e5a4ac7fabc58f/package.json#L30) command. The resulting TypeScript file is then committed to source control, so at the moment the app bundles a client for the **dev** UOWS. However, there may be differences between the dev and prod UOWS, such as web methos taking different arguments or returning different data. So, the dev client may not always work with the prod service.

I'm not too sure how to get around this. We could generate both clients and choose which one to use at runtime, or sync our releases with the UOWS so that there are no breaking changes. Either way, I think that should be done as a separate PR as this may be enough to solve the problem for now. Maybe we shouldn't close the parent issue yet though.

martin-trajanovski (approved):
This looks good to me!

jekabs-karklins:
I have a suggestion, I'll try to post today, please wait before merging.

jekabs-karklins (approved):

TCMeldrum:
As Jekabs pointed out in his [suggestion ](https://github.com/UserOfficeProject/user-office-backend/pull/402#pullrequestreview-747762379) we will configure the environment every time the backend is started just wanted to check that is ok with us?

mr-field (approved):

TCMeldrum:
Hi there we were getting an error when login in with a token on production that was related to rejectionSanitizer.ts It was a very small fix but just wanted to run it by you guys at the ESS before I merged as I made the change after you reviewed it.

jekabs-karklins (approved):
Looks good to me

jekabs-karklins:
> Hi there we were getting an error when login in with a token on production that was related to rejectionSanitizer.ts It was a very small fix but just wanted to run it by you guys at the ESS before I merged as I made the change after you reviewed it.

The fix is fine, although I'm a little curious as to why is the request is not having a response .

TCMeldrum:
I am not entirely sure, I think the rejection sanitizer was trying to access the rejection field that did not exist. I am not sure why checking the token tripped this bug but not other requests.

PR #173, opened 2021-02-11
SWAP-1371 Few tables have no Primary Keys

peter-asztalos:
## Description

This PR adds missing primary keys and also removes some unnecessary unique constraints.

## How Has This Been Tested

- [x] test environment database had no error after the patch
- [x] e2e test
- [x] manual

## Fixes

https://jira.esss.lu.se/browse/SWAP-1371

## Changes

- missing primary keys added

## Tests included/Docs Updated?

- [ ] I have added tests to cover my changes.
- [ ] All relevant doc has been updated

PR #400, opened 2021-09-06
#173 Display name instead of ID in toolbar

TeraMike:
## Description

When the external token is checked, a dummyUser is created as required for the database. This is saved as before, but now when returned this user gets the email, firstname, and lastname from the stfcUser added to it.  
I considered doing this in a datasource (so as to return a usable user from e.g.: ensureDummyUserExists) but couldn't find a neater way of doing it nor one that really fit the function -- open to ideas if there is a preferable alternative.

See frontend PR for screenshot.

## Fixes

Fixes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/173

## Depends on

PR for the toolbar changes in the frontend: https://github.com/UserOfficeProject/user-office-frontend/pull/541


TCMeldrum (approved):

ACLay (changes requested):
We actually have two first name fields in our users system, and it looks like this one does too.

In our current ones it's [`givenName` and `firstNameKnownAs`](https://github.com/isisbusapps/Users/blob/d6b401637298ceac6c1ea8af74380dc4c0f39d7a/users/users-model-ejb/src/main/java/com/stfc/UserOffice/dtos/PersonDetailsDTO.java#L18), and the here it's [`firstname` and `preferredname`](https://github.com/UserOfficeProject/user-office-backend/blob/89f7d72f6dcc3f4c254a4e2e1280de40f024d96a/src/resolvers/types/User.ts#L53)

Could you set that extra field on the dummyUser, and have the frontend use preferredname if it's been set instead of firstname?

TeraMike:
> In our current ones it's [`givenName` and `firstNameKnownAs`](https://github.com/isisbusapps/Users/blob/d6b401637298ceac6c1ea8af74380dc4c0f39d7a/users/users-model-ejb/src/main/java/com/stfc/UserOffice/dtos/PersonDetailsDTO.java#L18), and the here it's [`firstname` and `preferredname`](https://github.com/UserOfficeProject/user-office-backend/blob/89f7d72f6dcc3f4c254a4e2e1280de40f024d96a/src/resolvers/types/User.ts#L53)


Oh that's a good shout, I completely missed those since the different names just didn't register as the same thing in my head. I'll do that.

ACLay (approved):

src/mutations/UserMutations.ts
```diff
@@ -432,6 +432,12 @@ export default class UserMutations {
       const dummyUser = await this.dataSource.ensureDummyUserExists(userNumber);
       const roles = await this.dataSource.getUserRoles(dummyUser.id);
 
+      // add user info to dummmyUser for UserContext
```
TickleThePanda, line 435: Will this persist the data to the database? I'm guessing not but can we make the comment here clear to say that we shouldn't?

We should avoid storing any additional personal data in this database as we manage databases with personal data slightly different.

TickleThePanda (changes requested):

src/mutations/UserMutations.ts
TeraMike: No, this shouldn't persist the data since the dummyUser is created / written to database in the `ensureDummyUserExists` rabbithole above, which I haven't changed. This is done on login, checking the external token in `ExternalAuth`, which then handles login from the `UserContextProvider` file.

I'll change the comment to make this more clear.

TeraMike: @TickleThePanda does this look alright now?

TickleThePanda (approved):

PR #402, opened 2021-09-06
Simplify by moving related changes into one config file

jekabs-karklins:
## Description

This PR aims to simplify the logic of just having one mapping ConfigureEnvironment, and based on facility and environment the respective function is mapped


jekabs-karklins:
@TCMeldrum Please feel free to reject the suggestion as I'm not completely sure how is STFC site is supposed to work. 
This PR is just trying to simplify and reduce the amount of code by having only one configureEnvironment function.

For ESS it's mapping the function only in production but for STFC in both environments, as you mentioned you prefer.

As a side note I will just mention that if you run the function also in production, this means that that function will be run every time a new backend container is created, which might happen during scaling up as well.

src/config/dependencyConfigE2E.ts
```diff
@@ -26,11 +26,12 @@ import { SampleAuthorization } from '../utils/SampleAuthorization';
 import { ShipmentAuthorization } from '../utils/ShipmentAuthorization';
 import { UserAuthorization } from '../utils/UserAuthorization';
 import { VisitAuthorization } from '../utils/VisitAuthorization';
-import enableDefaultEssFeatures from './ess/enableDefaultEssFeatures';
-import setEssColourTheme from './ess/setEssColourTheme';
+import { configureESSDevelopmentEnvironment } from './ess/configureESSEnvironment';
 import { Tokens } from './Tokens';
 import { mapClass, mapValue } from './utils';
 
+const isProduction = process.env.NODE_ENV !== 'production';
```
TCMeldrum, line 33: Look like this isProduction isn't needed

TCMeldrum (commented):
Looks good thank you simplify nicely, I'll bring up that it is re-run on every startup in the other PR thread. Thanks, again!

TCMeldrum (approved):

PR #411, opened 2021-09-21
Swap 1719 try to exclude gateway conditionaly

martin-trajanovski:

### DIFF
--- a/db_patches/0106_AddScheduledEventsTable.sql
+++ b/db_patches/0106_AddScheduledEventsTable.sql
@@ -1,0 +1,20 @@
+DO
+$$
+BEGIN
+    IF register_patch('AddScheduledEventsTable.sql', 'Martin Trajanovski', 'Add scheduled_events table to store the scheduler information needed for the core frontend and remove the dependency', '2021-09-20') THEN
+
+    CREATE TABLE IF NOT EXISTS "scheduled_events" (
+        "scheduled_event_id" int NOT NULL
+      , "booking_type" varchar(30) NOT NULL
+      , "starts_at" TIMESTAMP NOT NULL
+      , "ends_at" TIMESTAMP NOT NULL
+      , "proposal_booking_id" int NOT NULL
+      , "proposal_pk" int NOT NULL
+      , "status" varchar(30) NOT NULL
+      , PRIMARY KEY (scheduled_event_id, proposal_booking_id)
+    );
+
+    END IF;
+END;
+$$
+LANGUAGE plpgsql;
--- a/src/datasources/mockups/ProposalDataSource.ts
+++ b/src/datasources/mockups/ProposalDataSource.ts
@@ -7,8 +7,15 @@ import {
   ProposalPksWithNextStatus,
 } from '../../models/Proposal';
 import { ProposalView } from '../../models/ProposalView';
+import { ScheduledEventCore } from '../../models/ScheduledEventCore';
 import { SepMeetingDecision } from '../../models/SepMeetingDecision';
 import { UpdateTechnicalReviewAssigneeInput } from '../../resolvers/mutations/UpdateTechnicalReviewAssignee';
+import {
+  ProposalBookingFilter,
+  ProposalBookingScheduledEventFilter,
+  ProposalBookingStatus,
+  ScheduledEventBookingType,
+} from '../../resolvers/types/ProposalBooking';
 import { ProposalEventsRecord } from '../postgres/records';
 import { ProposalDataSource } from '../ProposalDataSource';
 import { ProposalsFilter } from './../../resolvers/queries/ProposalsQuery';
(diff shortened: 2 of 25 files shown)


src/datasources/mockups/ProposalDataSource.ts
```diff
@@ -254,4 +271,36 @@ export class ProposalDataSourceMock implements ProposalDataSource {
...
+    proposalBookingId: number,
+    filter?: ProposalBookingScheduledEventFilterCore
+  ): Promise<ScheduledEventCore[] | null> {
+    return [dummyScheduledEventCore];
+  }
+
+  async addProposalBookingScheduledEvent(
+    eventMessage: ScheduledEventCore
+  ): Promise<void> {
+    // Not implemented
```
jekabs-karklins, line 292: Don't we want unit tests coverage as well, or could that be done in a separate task?

jekabs-karklins (approved):

src/datasources/mockups/ProposalDataSource.ts
martin-trajanovski: Yes that could be a task for the scheduler because it has no unit tests at all since it is started. If we want to add some we will need more time setting everything needed.

martin-trajanovski: We can definitely have unit tests to cover this as well.

TickleThePanda:
Sorry for my delay on this one. I'm asking somebody from P&O to review this instead of my as they will likely have a better understanding of the impact.

TickleThePanda (approved):
At a surface understanding of the changes, I'm happy with this.

martin-trajanovski:
> At a surface understanding of the changes, I'm happy with this.

Yes maybe we should try it and see how it works for all of us but I hope will be an improvement for sure.

db_patches/0106_AddScheduledEventsTable.sql
```diff
@@ -0,0 +1,20 @@
...
+BEGIN
+    IF register_patch('AddScheduledEventsTable.sql', 'Martin Trajanovski', 'Add scheduled_events table to store the scheduler information needed for the core frontend and remove the dependency', '2021-09-20') THEN
+
+    CREATE TABLE IF NOT EXISTS "scheduled_events" (
+        "scheduled_event_id" int NOT NULL
+      , "booking_type" varchar(30) NOT NULL
+      , "starts_at" TIMESTAMP NOT NULL
+      , "ends_at" TIMESTAMP NOT NULL
+      , "proposal_booking_id" int NOT NULL
+      , "proposal_pk" int NOT NULL
```
mr-field, line 12: Should this be a foreign key pointing to the proposal table? That would ensure that every event is linked to an existing proposal

src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -712,4 +721,124 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
...
+      'scheduled_events'
+    )
+      .select()
+      .where('proposal_pk', proposalPk)
+      .modify((qb) => {
+        if (filter?.status) {
+          qb.whereIn('status', filter.status);
+        }
+      })
+      .first();
```
mr-field, line 739: From what I understand, it's safe to return the first result here because all events linked to the same proposal will have the same booking ID. However, there's no guarantee in the (core) DB that will be the case; could we add something like a `UNIQUE` constraint to ensure that the same `proposal_pk` will always be linked to the same `proposal_booking_id`?

src/datasources/mockups/ProposalDataSource.ts
```diff
@@ -254,4 +271,36 @@ export class ProposalDataSourceMock implements ProposalDataSource {
...
+    eventMessage: ScheduledEventCore[]
+  ): Promise<void> {
+    return;
+  }
+
+  async updateProposalBookingScheduledEvent(
+    eventMessage: ScheduledEventCore
+  ): Promise<void> {
+    return;
+  }
```
mr-field, line 305: As Jekabs said, it would be good to have unit tests covering these as well. Could they be added in this PR?

martin-trajanovski: I think I couldn't add unit tests because there is no functionality to test here. Those are not mutations just functions to modify the database records. The only thing that could be added are integration tests but I am not sure if we execute them anywhere in the flow.

src/datasources/postgres/ProposalDataSource.ts
martin-trajanovski: I think we can add some constraints at least I can try to do it.

db_patches/0106_AddScheduledEventsTable.sql
```diff
@@ -1,20 +1,51 @@
...
+      t_row scheduled_events%rowtype;
+      BEGIN
+        FOR 
+          t_row IN SELECT * FROM scheduled_events
+          WHERE proposal_booking_id = new.proposal_booking_id
+          OR proposal_pk = new.proposal_pk
+        LOOP
+          IF 
+            (new.proposal_booking_id = t_row.proposal_booking_id AND new.proposal_pk = t_row.proposal_pk) OR
+            (new.proposal_booking_id != t_row.proposal_booking_id AND new.proposal_pk != t_row.proposal_pk)
```
mr-field, line 33: Minor point, but can this condition even happen? We're looping over all rows where either the `proposal_booking_id` or the `proposal_pk` matches `new`, so can both field be different at the same time?

martin-trajanovski: Or maybe you are right. Because I am querying only existing ones so there is no need to check if they are different that will be added right after the loop with RETURN NEW. I will remove this and test but it should work. Thanks for the good tip.

PR #421, opened 2021-09-30
SWAP-1853-upgrade-typescript-to-latest

martin-trajanovski:
## Description

refactor: Upgrade Typescript, Knex.js and all other related packages to latest.

## Motivation and Context

For some time we were experiencing some errors in the editors and we couldn't upgrade knex.js mostly because of typescript version that we were using. Now all the errors should be gone and everything should be up to date.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1853
https://jira.esss.lu.se/browse/SWAP-1851

## Depends on

https://github.com/UserOfficeProject/user-office-lib/pull/71


martin-trajanovski:
This https://github.com/UserOfficeProject/user-office-lib/pull/71 should be merged first and then logger version should be updated for this to work.

src/mutations/CallMutations.ts
```diff
@@ -49,7 +49,7 @@ export default class CallMutations {
       const result = await this.dataSource.delete(callId);
 
       return result;
-    } catch (error) {
+    } catch (error: any) {
```
jekabs-karklins, line 52: I would suggest like this, then we keep error as `unknown` for the rest of the clause, and cast it to whatever right before doing something with it.
```
} catch (error) {
      if ((error as {code:string}).code === '23503') {
```

or if there are issues with it then I think we can simplify 
`if ('code' in error && error.code === '23503')`
to
`if (error.code === '23503')`

src/utils/UserAuthorization.ts
```diff
@@ -60,14 +60,6 @@ export class UserAuthorization {
     }
   }
 
-  async isMemberOfProposal(
```
jekabs-karklins, line 63: This should be kept in, to properly display function overload
With this the code hint is proper
image

Without it
image

src/utils/VisitAuthorization.ts
```diff
@@ -105,7 +105,7 @@ export class VisitAuthorization {
      * User can modify the visit if he is a participant of a proposal
      * and the visit is not yet accepted
      */
-    if (this.userAuthorization.isMemberOfProposal(agent, proposal)) {
+    if (await this.userAuthorization.isMemberOfProposal(agent, proposal)) {
```
jekabs-karklins, line 108: 👍🏻  Good catch

src/utils/UserAuthorization.ts
martin-trajanovski: @jekabs-karklins I don't understand this. Not sure how with that hints are better than when it is defined as it should be only once. I see that second parameter is `proposalOrPk` which can be `Proposal | number | null` not sure why we need to define the function types three times.

jekabs-karklins: It is function overload syntax for TypeScript
https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

The same method can be invoked with different signature.

jekabs-karklins (approved):

PR #429, opened 2021-10-11
refactor: authorizers refactoring

jekabs-karklins:
## Description

This PR refactors authorizers.
Until now authorizers were mapped to implementation in dependencyConfig, but they were all mapped to the same implementation. As there is no need to map to different implementations, the authorizers can simply be resolved from the container, before they are used.

This PR also makes naming of authorizers consistent.

In case in the future the authorization logic is different in some other facility, then the authorizer mapping can be moved back to the dependencyConfig.


martin-trajanovski (approved):
I like this new approach! It seems to be much simpler and cleans the whole solution a bit. 🥇

PR #434, opened 2021-10-13
refactor: Integrate ESI with ScheduledEventId (Swap 1879)

jekabs-karklins:
## Description

This PR aims incorporate the recent change in PR https://github.com/UserOfficeProject/user-office-backend/pull/411 Refactor for making core independent from scheduler. 
The abovementioned PR allows to reference scheduled event in visits and in experiment_safety_inputs table thus simplifying CreateVisit and UpdateVisit mutations as well as improving DataBase integrity

## Fixes

https://jira.esss.lu.se/browse/SWAP-1879

## Changes

- Added `ScheduledEventDataSource` datasource
- Updated `CreateVisit` mutation
- Update `UpdateVisit` mutation
- Changed ExperimentSafetyInput structure (decoupling from visits)
- Added DB constraints

## Depends on
https://github.com/UserOfficeProject/user-office-frontend/pull/591

### DIFF
--- a/src/eventHandlers/messageBroker.ts
+++ b/src/eventHandlers/messageBroker.ts
@@ -25,6 +25,27 @@ type ProposalMessageData = {
   proposer?: { firstName: string; lastName: string; email: string };
 };
 
+let rabbitMQCachedBroker: null | RabbitMQMessageBroker = null;
+
+const getRabbitMQMessageBroker = () => {
+  if (rabbitMQCachedBroker === null) {
+    rabbitMQCachedBroker = createRabbitMQMessageBroker();
+  }
+
+  return rabbitMQCachedBroker;
+};
+
+const createRabbitMQMessageBroker = () => {
+  const messageBroker = new RabbitMQMessageBroker();
+  messageBroker.setup({
+    hostname: process.env.RABBITMQ_HOSTNAME,
+    username: process.env.RABBITMQ_USERNAME,
+    password: process.env.RABBITMQ_PASSWORD,
+  });
+
+  return messageBroker;
+};
+
 export function createPostToQueueHandler() {
   // return the mapped implementation
   return container.resolve<EventHandler<ApplicationEvent>>(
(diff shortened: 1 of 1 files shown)


db_patches/0109_ESIScheduledEventIdConstraint.sql
```diff
@@ -0,0 +1,50 @@
+DO
+$$
+BEGIN
+    IF register_patch('ESIScheduledEventIdConstraint.sql', 'Jekabs Karklins', 'Referecing scheduled_event_id from visits table', '2021-10-13') THEN
+
```
martin-trajanovski, line 5: I hope all migrations here are tested with some test data and they wont bring some problems when run into the test server.

src/datasources/postgres/ScheduledEventDataSource.ts
```diff
@@ -0,0 +1,19 @@
+import { ScheduledEventCore } from '../../models/ScheduledEventCore';
+import { ScheduledEventDataSource } from '../ScheduledEventDataSource';
+import database from './database';
+import { createScheduledEventObject, ScheduledEventRecord } from './records';
+
+export default class PostgresScheduledEventDataSource
```
martin-trajanovski, line 6: Maybe we need to move all other methods like (`proposalBookingScheduledEvents, addProposalBookingScheduledEvent, updateProposalBookingScheduledEvent and removeProposalBookingScheduledEvents`) from `ProposalDataSource.ts` because they have to be in the same file. They are all about scheduled events but I didn't created new datasource file at that point when added those methods but instead I put them in the ProposalDataSource.ts. Maybe this should be another task as a small refactor.

db_patches/0109_ESIScheduledEventIdConstraint.sql
jekabs-karklins: - [x]  Tested with test server data

PR #446, opened 2021-10-19
Added STFC logout

TCMeldrum:
Closes: UserOfficeProject/stfc-user-office-project#219
## Description

Added the Session-Id as an optional component to AuthJwtPayload
Added resolver to log out an stfc user by their session ID. 
Move the external token logic into the data source with no-op for no stfc modes

## Motivation and Context

Currently clicking the logout button while in stfc mode does not log you out

## How Has This Been Tested

Manual testing

## Changes

-Added externalTokenLogout resolver
-Added sessionId if exists to the AuthJwtPayload
-Moved external token logic in userDataSource

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/601


TickleThePanda (changes requested):
This is mostly a wider review of what you're doing here.

I'm not sure that `UserMutations` is the best place for the logout logic and I think we might be able to have the frontend always do a logout action and then the backend can decide what to do with that logout.

There's a few options I can think of:
1. Move the logic to the user data sources and have no-op for the ESS and delegate to the UOWS for us. We could have the frontend do the same for us and the ESS for both of these rather than letting the frontend know further about the logout source.
2. Another option might be to have a `LogoutProvider` and then have a `NoOpLogoutProvider` for the ESS and `StfcExternalLogoutProvider` for us. This might reduce some of the logic on the frontend as well so that there's always a logout call to the backend.

For both of these cases, the examples can be configured in the dependency injection.

TCMeldrum:
Unless anyone has any preferences I'm going to create a new datasource `AuthenticationDatasourse` to conform to the current pattern in the backend.

src/mutations/UserMutations.ts
```diff
@@ -458,6 +460,30 @@ export default class UserMutations {
     }
   }
 
+  async logout(externalToken: string): Promise<string | Rejection> {
+    try {
+      console.log(externalToken);
```
mr-field, line 465: I don't think we should log every token we receive, it would be best to log this kind of data only if there was an error. Also, we should avoid using `console.log()` for logging and use the logger from `@esss-swap/duo-logger` instead (it automatically writes timestamps for each line and can log at different severity levels)

src/mutations/UserMutations.ts
```diff
@@ -458,6 +460,30 @@ export default class UserMutations {
...
+
+      const client = new UOWSSoapClient(process.env.UOWS_URL);
```
mr-field, line 467: If we're going to move this in `UserDataSource`, it may be possible to create the client only once and store it in a "class" attribute. That way we don't have to recreate one for every request, which could be expensive

src/mutations/UserMutations.ts
```diff
@@ -458,6 +460,30 @@ export default class UserMutations {
...
+      await client.logout(proposalsToken.externalToken);
+
+      //Check logout has worked
+      const rawStfcUser = await client.getPersonDetailsFromSessionId(
+        proposalsToken.externalToken
+      );
+
+      if (rawStfcUser) {
+        return rejection('Logout unsuccessful', { externalToken });
+      }
```
mr-field, line 479: I think the `logout` method would throw an exception or return some kind of error code if there was a problem. Can we just check that instead of calling the UOWS again?

src/mutations/UserMutations.ts
```diff
@@ -458,6 +460,30 @@ export default class UserMutations {
     }
   }
 
+  async logout(externalToken: string): Promise<string | Rejection> {
```
mr-field, line 463: From what I inderstand, the argument here is not the external token (i.e. our sessionId), but the full JWT for a user. If so, could this be renamed to something more descriptive (like just `token`, or maybe `jwt`)

src/datasources/stfc/UOWSSoapInterface.ts
```diff
@@ -2,16 +2,9 @@ import * as soap from 'soap';
 
 export default class UOWSSoapClient {
 
-private wsdlUrl: string;
+private wsdlUrl: string = process.env.UOWS_URL ?? 'https://devapis.facilities.rl.ac.uk/ws/UserOfficeWebService?wsdl';
```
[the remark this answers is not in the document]
TCMeldrum: Yeah, I changed this manually forgot this was auto-generated. Should I revert this change and open separate issue for updating this?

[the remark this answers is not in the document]
mr-field: Yes, that's probably the best option. Also you'll need to pass `process.env.UOWS_URL` to the client constructor again, as it won't be picked up automatically

TCMeldrum:
I have a slight problem with the backend tests. Since I moved the logic to the I have moved some tests from [UserMutations.spec.ts](https://github.com/UserOfficeProject/user-office-backend/blob/80a7036b2656d49de94638cf1fad7683e540b60e/src/mutations/UserMutations.spec.ts#L323-L336) tests to [StfcUserDataSource.spec.ts](https://github.com/UserOfficeProject/user-office-backend/blob/develop/src/datasources/stfc/StfcUserDataSource.spec.ts) but I can't find a way to cover a valid user being returned from a valid token. When a valid token is provided it makes a call to `ensureDummyUserExists` it which is a call to `PostgresUserDataSource` but it should be making a call to mock version instead. Does anyone have a better idea than `process.env.DEPENDENCY_CONFIG == 'test'` for changing to call to the mock datasource?

mr-field:
Can you use something like `jest.spyOn().mockImplementation()` like [here](https://github.com/UserOfficeProject/user-office-backend/blob/80a7036b2656d49de94638cf1fad7683e540b60e/src/datasources/stfc/StfcUserDataSource.spec.ts#L11)? It should let you overwrite `ensureDummyUserExists` with your own function, which could retuen a hardcoded object.

TCMeldrum:
I think that should work! Thanks, I haven't worked with jest before

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -390,4 +390,39 @@ export class StfcUserDataSource implements UserDataSource {
...
+    dummyUser.preferredname = stfcUser.firstNameKnownAs;
+    dummyUser.lastname = stfcUser.familyName;
+
+    return dummyUser;
+  }
+
+  async externalLogout(token: string): Promise<void> {
+    await client.logout(token);
+    const rawStfcUser = await client.getPersonDetailsFromSessionId(token);
+    console.log(rawStfcUser);
```
mr-field, line 421: I think this should be removed and replaced with a log message before we throw the error (like `Failed to log out user <user id>`). The message should be logged with the main `logger` rather than the console, so that it's timestamped. According to [our logging guidelines](https://github.com/isisbusapps/ISISBusApps/wiki/Logging#logging-levels), it should be logged at the `warn` level

src/mutations/UserMutations.ts
```diff
@@ -458,6 +442,18 @@ export default class UserMutations {
     }
   }
 
+  async externalLogout(token: string): Promise<void | Rejection> {
+    try {
+      const decodedToken = verifyToken<AuthJwtPayload>(token);
+
+      this.dataSource.externalLogout(decodedToken.externalToken ?? 'no-token');
```
mr-field, line 449: Does TypeScript complain if you don't coalesce `null` to `no-token`? If so, it might stop if you handle the case where `decodedToken.externalToken` is null before calling the function, with something like
```
if (!decodedToken.externalToken) {
  throw("No external token found in JWT")
}
```

I don't know how logging works in these cases, but if rejections aren't automatically logged we'd also want to log that an error happened

mr-field (commented):
Just two small niggles left, otherwise looks good

src/mutations/UserMutations.ts
TCMeldrum: I am hesitant to add a rejection or log here as it is called in both [stfc and ess](https://github.com/UserOfficeProject/user-office-frontend/blob/0402baf290506d715adea382e71204c5659b7fab/src/context/UserContextProvider.tsx#L174-L176) modes as panda suggested and I don't want to clog to the ESS's logs.

src/resolvers/mutations/LogoutMutation.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { ResolverContext } from '../../context';
+import { LogoutTokenWrap } from '../types/CommonWrappers';
+import { wrapResponse } from '../wrapResponse';
+
+@Resolver()
+export class LogoutMutation {
+  @Mutation(() => LogoutTokenWrap)
+  logout(
+    @Arg('externalToken') externalToken: string,
```
mr-field, line 11: The last step is to rename this to just `token` as it's the full JWT, not the external token from the UOWS

mr-field: The `externalLogout` method in `UserDataSource` should also be renamed to just `logout`, as the method is an abstract way of handling logout which just happens to be external in our implementation

mr-field (approved):

src/mutations/UserMutations.ts
```diff
@@ -46,10 +47,20 @@ import { UserAuthorization } from '../utils/UserAuthorization';
...
-    @inject(Tokens.UserDataSource) private dataSource: UserDataSource
-  ) {}
+    @inject(Tokens.UserDataSource) private dataSource: UserDataSource,
+    @inject(Tokens.AdminDataSource) private adminDataSource: AdminDataSource
+  ) {
+    adminDataSource.getFeatures().then((features) => {
+      this.externalAuth = features.filter(
+        (feature) => feature.id == FeatureId.EXTERNAL_AUTH
+      )[0].isEnabled;
+      console.log(this.externalAuth);
```
TCMeldrum, line 61: As a note, I just ran into a small bug with this. When this class is initialised the variable in the database might not have been sent to its true value yet as the `setStfcEnviroment` method in the config might not have run. This will only be a bug locally when doing `fba-compose down` + `fba-compose up`  as the database does not have to be rebuilt on dev or prod.

TickleThePanda (approved):

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -390,4 +392,40 @@ export class StfcUserDataSource implements UserDataSource {
   async getRoleByShortCode(roleShortCode: Roles): Promise<Role> {
     throw new Error('Method not implemented.');
   }
+
+  async checkExternalToken(token: string): Promise<User> {
```
ACLay, line 396: I realise it's just moving a method and maintaining existing functionality, but can we rename/refactor (or open an issue to do so for) `checkExternalToken`? With a name like that I'd expect it to return a boolean or some sort of report on the token's status, but it's actually fetching a user and throwing an error if it can't find one.

TCMeldrum: Sure I'll change that is `externalTokenLogin` better?

ACLay: Yeah, that sounds much better

ACLay (approved):

TCMeldrum:
@UserOfficeProject/swap Would you mind reviewing the backend part of this PR. I see you have done the frontend just wanted to check everything is OK on this side.

bolmsten (approved):

bolmsten:
@TCMeldrum we had a look at it and I think we should create a loginDependency injection to keep this logic outside datasources but that would require a bigger rewrite. I think it is fine to merge this now and get back to it later as you probably need this now for going live

TCMeldrum:
> @TCMeldrum we had a look at it and I think we should create a loginDependency injection to keep this logic outside datasources but that would require a bigger rewrite. I think it is fine to merge this now and get back to it later as you probably need this now for going live

Ok cool thanks I'll open an issue to refactor it after the release has gone out. Just to let you know there will be one more PR about making sure users stay logged into our current systems that I will be putting together shortly that will also probably need to be part of that refactoring. Any queries let me know

PR #462, opened 2021-11-03
Fix bug where Instrument Scientists cannot open proposals they don't own

IVlaD17:
## Description
Added an isInstrumentScientist check in `UserAuthorization.ts`

## Motivation and Context
Closes UserOfficeProject/stfc-user-office-project#256

## How Has This Been Tested
Manual testing throughout the front-end.


simonfernandes (commented):
It looks like all of the other `isRole()` methods check against the current role, rather than this more general check that the user has a role. I'm not sure what the implications of this are (maybe they can do things when logged in as a basic user that they shouldn't be able to do?), but I'd be minded to think we should probably follow how the other's have been done

IVlaD17:
> It looks like all of the other `isRole()` methods check against the current role, rather than this more general check that the user has a role. I'm not sure what the implications of this are (maybe they can do things when logged in as a basic user that they shouldn't be able to do?), but I'd be minded to think we should probably follow how the other's have been done

Like this?

simonfernandes (commented):
Looks good - could you just fix the whitespace problems with this? The ESS have asked us to ensure we lint our code. This is usually done automatically when we make commits and they'll be rejected if there's any errors, but I think @TCMeldrum mentioned this isn't currently working . So instead (or as well as) we can install the ESLint VS Code extension and it'll highlight and offer fixes as you write. There are some instructions on Confluence for this

simonfernandes (approved):

PR #467, opened 2021-11-05
fix: reference number integration tests

simonfernandes:
## Description

This fixes the reference number integration tests. 

They were failing (or not running) due to some refactoring of column names. 

The two removed tests are because they are no longer needed due to frontend validation.

## How Has This Been Tested

Tests pass, logged in and used proposal system

## Fixes

Closes UserOfficeProject/stfc-user-office-project#266

- [x] All relevant doc has been updated (issue created: https://github.com/UserOfficeProject/stfc-user-office-project/issues/267, STFC confluence updated, will post on slack)


mr-field (commented):
The updates look fine, but can we keep at least one of the removed tests to check what happens if the reference number format is invalid?

src/datasources/postgres/CallDataSource.test.ts
```diff
@@ -111,6 +117,8 @@ async function teardown() {
 
   await database('templates').where('template_id', -999).del();
 
+  await database('templates').where('template_id', -998).del();
+
```
mr-field, line 121: I just noticed, do we need both of these? The rest of the code looks fine

simonfernandes: I think so, on line 93 there's now this additional template being inserted, so this is making sure to delete them both

mr-field: Oh sorry, I wasn't wearing my glasses and read both IDs as `998`

mr-field (approved):

bolmsten (approved):

PR #472, opened 2021-11-10
fix: SMTP emails not sending

simonfernandes:
## Description

This fixes emails not being sent locally only (tested). 

I assume this will also be needed for the dev and prod emails which are also broken, but I'm unable to test that and it additionally requires https://github.com/isisbusapps/docker-orchestration/pull/185

I think this may have stopped working when we switched to Linux containers.

## How Has This Been Tested

Emails send locally only

## Fixes

Part of UserOfficeProject/stfc-user-office-project#233


src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -62,11 +62,10 @@ export class SMTPMailService extends MailService {
 
     if (
       !(await (this._email as any).templateExists(
-        options.content.template_id + '\\html.pug'
+        options.content.template_id + '/html.pug'
```
TickleThePanda, line 65: Could we use `path.join(options.content.template_id, 'html.pug')` here?

I think path can be imported as:
```
import * as path from 'path';
```

ACLay (changes requested):
To ensure that any future changes to the config path don't cause this to pop up again can you switch this to use a path handler library rather than just concatenating strings? It looks like we already have one in use elsewhere in the backend: https://github.com/UserOfficeProject/user-office-backend/blob/df93e842224e35ef9a987e54f66af073083aecaf/src/datasources/postgres/AdminDataSource.ts#L37

TickleThePanda (dismissed):

ACLay (approved):

PR #481, opened 2021-11-16
feat(feedback): Adding Feedback feature

jekabs-karklins:
## Description

This PR implements feature wich allows user to provide feedback once the experiment is done

## Motivation and Context
image

The feedback is just another template which can be configured in administration panel. The Feedback is attached to scheduled event. It is enabled once the experiment is completed.

## Fixes

https://jira.esss.lu.se/browse/SWAP-SWAP-1951

## Changes

- New template

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/648


src/models/Template.ts
```diff
@@ -96,10 +97,11 @@ export class TemplateCategory {
 
 export enum TemplateCategoryId {
   PROPOSAL_QUESTIONARY = 1,
-  SAMPLE_DECLARATION,
-  SHIPMENT_DECLARATION,
-  VISIT_REGISTRATION,
+  SAMPLE_DECLARATION = 2,
+  SHIPMENT_DECLARATION = 3,
+  VISIT_REGISTRATION = 4,
   GENERIC_TEMPLATE = 7,
```
martin-trajanovski, line 103: I don't understand why numbers are going from 4 to 7 or I am missing something here.

src/mutations/FeedbackMutations.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  @Authorized()
+  async deleteFeedback(
+    user: UserWithRole | null,
+    feedbackId: number
+  ): Promise<Feedback | Rejection> {
+    const hasRights = await this.feedbackAuth.hasWriteRights(user, feedbackId);
+    if (hasRights === false) {
+      return rejection(
+        'Can not update feedback because of insufficient permissions',
```
martin-trajanovski, line 171: Just a minor thing in the text should be deleted instead of updated

martin-trajanovski (approved):
Otherwise it looks great!

src/models/Template.ts
jekabs-karklins: Two categories were deleted during the refactoring,
I have a task to change this enum from numbers to strings, so it is more readable.

PR #496, opened 2021-11-24
style: alter log format

simonfernandes:
## Description

This is an optional PR that alters the event logger format to mirror the console logger format

Before (from STFC log server):
image

After (local log output):
image

## How Has This Been Tested

Tested by checking local log output

## Fixes

Part of UserOfficeProject/stfc-user-office-project#272

### DIFF
--- a/src/eventHandlers/logging.ts
+++ b/src/eventHandlers/logging.ts
@@ -14,8 +14,8 @@ export default function createHandler() {
   // Handler that logs every mutation wrapped with the event bus event to stdout and event_logs table.
   return async function loggingHandler(event: ApplicationEvent) {
     const json = JSON.stringify(event);
-    const timestamp = new Date().toLocaleString();
-    console.log(`${timestamp} -- ${json}`);
+    const timestamp = new Date().toISOString();
+    console.log(`[${timestamp}] EVENT - ${json}`);
 
     // NOTE: If the event is rejection than log that in the database as well. Later we will be able to see all errors that happened.
     if (event.isRejection) {


src/eventHandlers/logging.ts
```diff
@@ -14,8 +14,8 @@ export default function createHandler() {
   // Handler that logs every mutation wrapped with the event bus event to stdout and event_logs table.
   return async function loggingHandler(event: ApplicationEvent) {
     const json = JSON.stringify(event);
-    const timestamp = new Date().toLocaleString();
-    console.log(`${timestamp} -- ${json}`);
+    const timestamp = new Date().toISOString();
+    console.log(`[${timestamp}] EVENT - ${json}`);
```
TickleThePanda, line 18: Rather than using `console.log` here, would it be worth using `logger`? Are there any other places we're using `console.log` outside of tests?

TickleThePanda (changes requested):

src/eventHandlers/logging.ts
simonfernandes: I can try this. There's around 15 other instances of `console.log` in the frontend and backend.

TickleThePanda: It might be worth moving those to use the `console.log` if they're useful - or delete them instead.

simonfernandes: I've made this PR to handle events using the logger: https://github.com/UserOfficeProject/user-office-lib/pull/81. If that gets approved then I'll amend this PR to use `logger`.

Removing the remaining `console.log` statements is being handled by Cosimo as part of another issue

jekabs-karklins (approved):

mr-field (changes requested):
I've updated this file in https://github.com/UserOfficeProject/user-office-backend/pull/507 to log all events with `duo-logger`. This ensures they follow the same format as everything else.

If the new format is OK, I think this PR should be closed. If the log messages aren't clear yet, we should solve the merge conflict here and make changes to the new log statement.

simonfernandes:
Closing as no longer necessary after #507

PR #504, opened 2021-11-30
Refactor: Instrument sci query now uses view

TCMeldrum:
Part of: UserOfficeProject/stfc-user-office-project#254

## Description

Changed the Instrument scientists get proposals query to use the proposal view instead to cut down on the needed DB queries. Added a couple of columns to the view from existing tables to keep the functionality of the query.

## Motivation and Context

The instrument scientist query was very slow for a large number of proposals

## How Has This Been Tested

Tests all pass

## Fixes

UserOfficeProject/stfc-user-office-project#254

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/647


src/datasources/mockups/ProposalDataSource.ts
```diff
@@ -226,7 +226,7 @@ export class ProposalDataSourceMock implements ProposalDataSource {
     first?: number,
     offset?: number
   ) {
-    return { totalCount: 1, proposals: [dummyProposal] };
+    return { totalCount: 1, proposals: [] };
```
mr-field, line 229: I looks like this function is supposed to return at least one result. I'm guessing it's not actually used as the tests are passing, but could you add a dummy proposal with the new type?

src/resolvers/queries/ProposalsViewQuery.ts
```diff
@@ -8,6 +17,21 @@ import { ProposalsFilter } from './ProposalsQuery';
 class ProposalsViewArgs {
   @Field(() => ProposalsFilter, { nullable: true })
   public filter?: ProposalsFilter;
+
+  @Field(() => Int, { nullable: true })
+  public first?: number;
+
+  @Field(() => Int, { nullable: true })
+  public offset?: number;
```
mr-field, line 25: These arguments are only used by `instrumentScientistProposals()`, but they are also declared as the arguments of `proposalsView()`. Could you either add them only to `instrumentScientistProposals()`, or change `proposalsView()` so that it also uses them?

martin-trajanovski (approved):

mr-field (approved):

PR #510, opened 2021-12-08
Swap 2015 assign local contact to timeslot

martin-trajanovski:
## Description

Added beamlineManager resolver to the instrument to be able to query from the scheduler and localContact resolver to show it in the user experiments table.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2015

## Changes

https://user-images.githubusercontent.com/6333063/145542409-78c01e45-3d3e-4967-9370-7fee15087bea.mp4

## Depends on

https://github.com/UserOfficeProject/user-office-scheduler-backend/pull/106


db_patches/0114_AddLocalContactToScheduledEvents.sql
```diff
@@ -0,0 +1,15 @@
+DO
+$DO$
+BEGIN
+  IF register_patch('AddLocalContactToScheduledEvents.sql', 'Martin Trajanovski', 'Add local contact person to the scheduled event', '2021-12-08') THEN
+  BEGIN
+
+    ALTER TABLE "scheduled_events"
+    ADD COLUMN IF NOT EXISTS "local_contact" int DEFAULT NULL
+    REFERENCES users(user_id) ON DELETE CASCADE;
```
jekabs-karklins, line 9: I would suggest to not have here `ON DELETE CASCADE`. 
If someone tries to delete the user, let it fail instead of automatically removing scheduled events to whom the user is local contact.

Alternatively set the local_contact to null when deleting which also would work, as the default value is null already
`REFERENCES users(user_id) ON DELETE SET NULL`

src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -333,6 +345,7 @@ export default class PostgresInstrumentDataSource
         'name',
         'i.short_code',
         'description',
+        'i.manager_user_id',
```
jekabs-karklins, line 348: do we need to have `i.` here? I think this is only needed if there is conflicting names on both tables

jekabs-karklins (approved):

PR #512, opened 2021-12-13
Swap 2010 implement questionary import

jekabs-karklins:
## Description

This PR aims to add feature which allows exported questionaries to be imported

## Motivation and Context

This feature can be useful in many scenarios.
One scenario is where user offices could use test server to prepare the template and then export it and import in production environment.

## How Has This Been Tested

Added E2E

## Fixes

https://jira.esss.lu.se/browse/SWAP-2010

## Changes

- Increasing GQL request size to 5MB
- Implementing Validate template JSON
- Implementing Import template JSON

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/689


bolmsten (approved):

src/datasources/postgres/TemplateDataSource.ts
```diff
@@ -191,6 +208,125 @@ export default class PostgresTemplateDataSource implements TemplateDataSource {
...
+
+      if (!existingQuestion) {
+        questionComparisons.push({
+          existingQuestion: null,
+          newQuestion: newQuestion,
+          status: QuestionComparisonStatus.NEW,
+          conflictResolutionStrategy: ConflictResolutionStrategy.USE_NEW,
+        });
+      } else {
+        if (jsonDeepEqual(newQuestion, existingQuestion)) {
```
martin-trajanovski, line 297: I guess question properties could be in different order that's why you use this helper method instead of using just JSON.stringify on the objects and compare.

src/utils/json.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export default function jsonDeepEqual(a: any, b: any): boolean {
```
martin-trajanovski, line 2: You could probably do something like this here: `export default  function jsonDeepEqual<T>(a: T, b: T): boolean {` with generic types instead of using `any` and disabling the eslint rule.

martin-trajanovski (approved):
Just a small comment on the helper method types otherwise looks great!

src/datasources/postgres/TemplateDataSource.ts
jekabs-karklins: Yes, that was the case

PR #513, opened 2021-12-14
feat: Add country to institution 

bolmsten:
## Description

See: https://github.com/UserOfficeProject/user-office-frontend/pull/691


db_patches/0114_AddCountryToInstitutions.sql
```diff
@@ -0,0 +1,13 @@
+DO
+$$
+BEGIN
+    IF register_patch('AddCountryToInstitutions.sql', 'Fredrik Bolmsten', 'Add a country field for institutions, useful for KPI generation', '2021-12-06') THEN
+    ALTER TABLE "institutions" ADD "country_id" INT DEFAULT 1;
```
jekabs-karklins, line 5: Maybe you could add
`ALTER TABLE institutions ALTER COLUMN country_id DROP DEFAULT;`
because while setting 1 to existing institutions makes sense during migration, any new ones added should be defined.

jekabs-karklins (approved):

PR #518, opened 2021-12-15
Upgrade and remove unused dependencies to reduce vulnerabilities

martin-trajanovski:
## Description

Upgraded some of the dependencies and removed the unused ones to reduce the number of package vulnerabilities.

## Motivation and Context

Previously we had around 30 known package vulnerabilities now that is resolved.

## How Has This Been Tested

Created PR on the frontend corresponding to this branch to run e2e tests against this upgrade and it seems good so far.

## Changes

From this:
Screenshot_7
to this:
Screenshot_8


src/utils/buildFederatedSchema.ts
```diff
@@ -25,13 +28,23 @@ export async function buildFederatedSchema(
...
   });
 
+  const env = process.env.NODE_ENV;
+
+  // NOTE: applyMiddleware must be before addResolversToSchema as a workaround for the issue: https://github.com/maticzav/graphql-middleware/issues/395
+  if (env === 'production') {
+    // prevent exposing too much information when running in production
+    federatedSchema = applyMiddleware(schema, rejectionSanitizer);
+  } else {
+    federatedSchema = applyMiddleware(federatedSchema, rejectionLogger);
```
mr-field, line 45: Looks like the first call to `applyMiddleware()` passes `schema`, while the second uses `federatedSchema`. Is that intentional? I think we used to pass the same object on both environments.

martin-trajanovski: You are right. I will double check this I think it is by mistake. Thanks

mr-field (approved):

jekabs-karklins (approved):

PR #537, opened 2022-01-11
Pin Faker version to 5.5.3

mr-field:
Part of UserOfficeProject/stfc-user-office-project#312

## Description
This prevents us from picking up any future updates, now that the package has been compromised.

NPM packages are immutable once uploaded, so as long as the faker package isn't completely deleted we should be safe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "eslint-plugin-jest": "^24.5.0",
     "eslint-plugin-prettier": "^3.4.1",
     "eslint-plugin-unused-imports": "^1.1.4",
-    "faker": "^5.5.3",
+    "faker": "5.5.3",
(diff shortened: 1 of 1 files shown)


jekabs-karklins (approved):
👍🏻

package.json
```diff
@@ -117,7 +117,7 @@
     "eslint-plugin-jest": "^24.5.0",
     "eslint-plugin-prettier": "^3.4.1",
     "eslint-plugin-unused-imports": "^1.1.4",
-    "faker": "^5.5.3",
+    "faker": "5.5.3",
```
martin-trajanovski, line 120: Putting a note also here might be worth not forgetting this in the future.

martin-trajanovski (approved):

PR #538, opened 2022-01-13
Hotifx so that email invites get associated orcid

bolmsten:
## Description

At the moment a user receiving an email invite will not get a associated orcid on sign up with their account as the orcid arg is removed in the update user function of userMutations, this PR instead goes directly to the datasource when finding that a user is a placeholder 


src/mutations/UserMutations.ts
```diff
@@ -203,11 +203,17 @@ export default class UserMutations {
...
+        .update({
+          ...user,
+          ...args,
+        })
+        .then((user) => user as UserWithRole)
+        .catch((err) => {
+          return rejection('Could not update user', { user }, err);
+        });
+
       if (isRejection(updatedUser) || !changePassword) {
```
jekabs-karklins, line 217: This can never be rejection.
Because the signature of datasource is 
`update(user: User): Promise<User>;`

jekabs-karklins (approved):
Looks good, except since now we are calling DataSource the signature of the response object is different

src/mutations/UserMutations.ts
jekabs-karklins: Hi, I am referring to the line 217.

`if (isRejection(updatedUser) || !changePassword) {`

the `isRejection(updatedUser)` will always be `false`

PR #541, opened 2022-01-13
fix(shipments): Add ability to declare multiple shipments per experiment time

jekabs-karklins:
## Description

This PR aims to address the new/revised requirements for how shipments should be declared.

## Motivation and Context
Now it is possible to add multiple shipments per experiment time.
getScheduled event is renamed to GetScheduledEventCore, to avoid naming conflicts

## Fixes

https://jira.esss.lu.se/browse/SWAP-2139

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/727


src/mutations/ScheduledEventMutations.ts
```diff
@@ -0,0 +1,19 @@
...
+@injectable()
+export default class ScheduledEventMutations {
+  private shipmentAuth = container.resolve(ShipmentAuthorization);
+  private userAuth = container.resolve(UserAuthorization);
+  constructor(
+    @inject(Tokens.ScheduledEventDataSource)
+    private scheduledEventDataSource: ScheduledEventDataSource,
+    @inject(Tokens.ShipmentDataSource)
+    private shipmentDataSource: ShipmentDataSource
+  ) {}
```
martin-trajanovski, line 18: Not sure why we need this file when there are no mutations or it needs to be there?

martin-trajanovski (approved):
Looks good to me! 👀 👍

PR #542, opened 2022-01-14
fix(scheduled-event): renaming scheduledEvent to scheduledEventCore

jekabs-karklins:
## Description

This is to avoid conflicts in naming between core and scheduler

## Motivation and Context

When using Apollo Federation, endpoint names must not clash.
There was a clashing query named ScheduledEvent


martin-trajanovski (approved):
Looks good just don't forget to update the sdk on the frontend.

PR #550, opened 2022-01-21
fix: As a developer i want meaning full logs (Swap-2197)

jekabs-karklins:
## Description

Making error logs more meaningful

## Motivation and Context

Because we are more in production now, it is important to gather good logs.

## Fixes

https://jira.esss.lu.se/browse/SWAP-2197

## Changes

- Integrated duo-logger 2
- Made sure there are no conflicting error message names
- Simplified logging for EAM service
- When SQL query happens, log with name "QUERY ERROR" is sent and now populating the field QueryName with SQL that went wrong. Latter this can be aggregated to provide meaningful insights into database errors.

## Depends on

N/A


martin-trajanovski (approved):
Looks quite good!

PR #553, opened 2022-01-25
Fixed the bug where User Officers could not add co-investigators to proposals

IVlaD17:
## Description
Changes the STFC `getUsers` method so that it can now pull Users from the UOWS back-end.

## Motivation and Context
Closes UserOfficeProject/stfc-user-office-project#284

## How Has This Been Tested
Manual testing in the front-end.

## Depends on
UserOfficeProject/user-office-frontend#674


jekabs-karklins (approved):

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -315,6 +315,20 @@ export class StfcUserDataSource implements UserDataSource {
...
+
+      const stfcBasicPeopleByLastName: StfcBasicPersonDetails[] | null = (
+        await client.getBasicPeopleDetailsFromSurname(token, filter, true)
+      )?.return;
+
+      users = stfcBasicPeopleByLastName
+        ? stfcBasicPeopleByLastName.map((person) =>
+            toEssBasicUserDetails(person)
+          )
+        : [];
```
mr-field, line 329: Looks like the UOWS returns an empty list rather than `null` if it didn't find any results, so we should be able to skip this check: https://github.com/isisbusapps/Users/blob/bd3236cd5c8d804eb39c1e2b40951f53e4d2cd61/users/users-model-ejb/src/main/java/com/stfc/UserOffice/model/basicpersondetails/BasicPersonDetailsService.java#L210

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -294,10 +294,10 @@ export class StfcUserDataSource implements UserDataSource {
   ): Promise<{ totalCount: number; users: BasicUserDetails[] }> {
     const dbUsers: BasicUserDetails[] = (
       await postgresUserDataSource.getUsers(
-        filter,
+        undefined,
         first,
         offset,
-        userRole,
+        undefined,
```
mr-field, line 300: For future reference, this will show all users even in dialogs which should be restricted to Instrument Scientists or User Officers. I think this is OK for now, as we don't have a way of getting the roles of many users in a single call. Those dialogs are currently broken, as our roles come from the UOWS and are not set in the database: the `userRole` filter always fails as it's looking for nonexisting data.

mr-field (commented):
In the future, it may be a good idea to add a method to the UOWS for getting the roles of a list of people. That would allow us to filter by user role, which we can't do at the moment.

src/datasources/stfc/StfcUserDataSource.ts
mr-field: Also, we should either do the `if (filter)` check early and return the data from the UOWS without running the DB query if we have a filter, or use `if/else` to either make the UOWS call or the DB query. Currently, the code always does the DB query and then thows away the result if there is a filter, which is not very efficient.

mr-field: Also, we should either do the `if (filter)` check early and return the data from the UOWS without running the DB query if we have a filter, or use `if/else` to either make the UOWS call or the DB query. Currently, the code always does the DB query and then thows away the result if there is a filter, which is not very efficient.

TCMeldrum: Cool updated do the if filter earlier and removed the check

src/datasources/stfc/StfcUserDataSource.ts
```diff
@@ -292,32 +292,50 @@ export class StfcUserDataSource implements UserDataSource {
...
-      users = stfcBasicPeople
-        ? stfcBasicPeople.map((person) => toEssBasicUserDetails(person))
-        : [];
+      userDetails = stfcBasicPeopleByLastName.map((person) =>
+        toEssBasicUserDetails(person)
+      );
+
+      finalTotalCount = userDetails.length;
+      // eslint-disable-next-line no-console
+      console.log(finalTotalCount);
```
mr-field, line 311: We should either log this with `duo-logger` and a clear `INFO` message, or remove it. Using the console directly doesn't record the timestamp or severity of the message, which is why we enabled to Eslint rule to ban `console` calls. `duo-logger` logs to the console anyway, so there are no downsides to using it even for debug messages.

mr-field (changes requested):
Just a likely leftover debug message left, otherwise looks good.

src/datasources/stfc/StfcUserDataSource.ts
TCMeldrum: Yeah, leftover debug message sorry

mr-field (approved):

PR #555, opened 2022-01-31
fix: Send revised fields to EAM

jekabs-karklins:
## Description

This PR aims to add new fields that are send to EAM, which they have recently implemented

## Motivation and Context
New fields added

- is_dangerous_goods
- dangerous_goods_un_number
- dangerous_goods_details
- shipment_sample_risks
- parcel_value
- shipment_sender_company
- shipment_sender_street_address
- shipment_sender_zip_code
- shipment_sender_city_country
- shipment_sender_name
- shipment_sender_email
- shipment_sender_phone

## Fixes

https://jira.esss.lu.se/browse/SWAP-2249

## Changes

EAM service


src/queries/ScheduledEventQueries.ts
```diff
@@ -26,4 +26,18 @@ export default class ScheduledEventQueries {
 
     return scheduledEvents;
   }
+
+  @Authorized([Roles.USER])
+  async getScheduledEventCore(
+    user: UserWithRole | null,
+    scheduledEventId: number
+  ): Promise<ScheduledEventCore | null> {
+    const scheduledEvents = await this.dataSource.getScheduledEventCore(
```
martin-trajanovski, line 35: Probably you should name the variable just `scheduledEvent` instead of using plural.

martin-trajanovski (approved):
Looks good.

PR #556, opened 2022-01-31
fix: Questions using MCQ as a dependency not showing in PDF

vyshnavi1605:
## Description

Closes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/324
## Motivation and Context

Any answer to a question that uses a multiple choice question as a dependency, does not show in the downloaded PDF. When downloading the PDF the method `isSatisfied()` is called and the statement `answer.value === params` returns false. Even when they both do hold the same value.
## How Has This Been Tested

Tested through creating a proposal and downloading it to check if all the questions and answers that should appear are appearing. 

PDF before changes were made:
image

PDF after the changes were made:
image

## Fixes

Fixes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/324
## Changes

File changed: `ConditionEvaluator.ts`


TickleThePanda (commented):
I'm not sure if this is idiomatic TS but if it is, I'm happy with it.

Are there any unit tests for this module? Could we add a new case here?

Once it is ready, it would be good to release it as a hotfix.

vyshnavi1605:
> I'm not sure if this is idiomatic TS but if it is, I'm happy with it.

I've done some research to double check but I can't seem to find a definite answer. `answer.value` seems to `['yes']` and `params` is `'yes'` I assume that's why false was being returned previously for the `isSatisfied` method. Converting both of the values to string and getting the primitive value seems to be working. `JSON.stringify` didn't seem to be working either.

vyshnavi1605:
> Are there any unit tests for this module? Could we add a new case here?

Yes the unit test for this is [here](https://github.com/UserOfficeProject/user-office-backend/blob/develop/src/models/ProposalModelFunctions.spec.ts#L47 )

jekabs-karklins (approved):

TickleThePanda:
I think the unit test could do with a new case when the values checked are MCQ. Once that's done, I think anybody else can re-review.

mr-field (approved):

jekabs-karklins (approved):

PR #557, opened 2022-02-03
feat(units): improve units handling

jekabs-karklins:
## Description

This PR aims to add feature that handles ambiguity when it comes to entering values with different units.

## Motivation and Context

Users have option to enter values in different units. For example there could be a question asking for length in centimeters and millimeters. To be able to search and find all answers greater then some value, additionally to original answer we store answer in SI units

## How Has This Been Tested

- [x] E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1656

## Changes

- Number Input
- Interval Input

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/743


db_patches/0117_MigrateUnits.sql
```diff
@@ -0,0 +1,281 @@
+DO
+$DO$
+BEGIN
+  IF register_patch('MigrateAnswersNumberInput.sql', 'Jekabs Karklins', 'Associate shipment with scheduled event', '2022-01-05') THEN
```
martin-trajanovski, line 4: Just a small note to change the description and the date for the patch. And also I hope you already tested this migration with some test data because it seems quite complex to me.

docker-compose.yml
```diff
@@ -30,6 +30,8 @@ services:
       - 5432:5432
     volumes:
       - ./db_patches/:/docker-entrypoint-initdb.d/
+    extra_hosts:
```
martin-trajanovski, line 33: Not sure why this extra_hosts is needed with the gateway.

martin-trajanovski (approved):
Maybe just small changes needed but otherwise looks good.

db_patches/0117_MigrateUnits.sql
jekabs-karklins: Yes this has been tested with production data.
Description has been updated. Thanks for spotting that.

docker-compose.yml
jekabs-karklins: This was accidental. Thanks for noticing that.

PR #562, opened 2022-02-11
feat(units): Import/export units

jekabs-karklins:
## Description

This PR aims to implement a feature to be able to transfer units from one environment to another

## Motivation and Context

Importing/exporting units makes it easier to move units around from one environment to another

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2128

## Changes

- new Units datasource
- new Units mutations/queries

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/759


src/datasources/postgres/UnitDataSource.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    await Promise.all(
+      units.map(async (unit) => {
+        const conflictResolution = conflictResolutions.find(
+          (resolution) => resolution.itemId === unit.id
+        );
+        switch (conflictResolution?.strategy) {
+          case ConflictResolutionStrategy.USE_NEW:
+            await this.deleteUnit(unit.id);
+            await this.createUnit(unit);
```
martin-trajanovski, line 236: Not quite sure what happens here if for example `deleteUnit` is successful but `createUnit` fails for whatever reason. There might be some better way of solving this using transactions but I am not fully aware of the logic and can't say much more.

src/queries/UnitQueries.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Tokens } from '../config/Tokens';
+import { UnitDataSource } from '../datasources/UnitDataSource';
+
+@injectable()
+export default class UnitQueries {
+  constructor(
+    @inject(Tokens.UnitDataSource) private unitDataSource: UnitDataSource
+  ) {}
+
+  async getUnits() {
```
martin-trajanovski, line 12: Is there any authorization check needed here for the queries or they should be open for now?

src/resolvers/types/CommonWrappers.ts
```diff
@@ -309,7 +317,7 @@ export class ApiAccessTokenResponseWrap extends ResponseWrapBase {
 @ObjectType()
 export class ProposalsResponseWrap extends ResponseWrapBase {
   @Response()
-  @Field(() => [Proposal])
+  @Field(() => [Proposal], { nullable: true })
```
martin-trajanovski, line 320: Is there any specific reason why the proposal response becomes nullable in this PR or it is a fix for something else?

martin-trajanovski (commented):
Looks good! I just left few comments for you to check.

TickleThePanda (approved):

src/datasources/postgres/UnitDataSource.ts
jekabs-karklins: I agree. We have multiple places like this. I think we need some generic solution to address this class of problems. I will create an investigation task in Jira https://jira.esss.lu.se/browse/SWAP-2400

src/queries/UnitQueries.ts
jekabs-karklins: There is no benefit for the authorizaiton here, i guess

src/resolvers/types/CommonWrappers.ts
jekabs-karklins: Yes, I noticed that in case mutation fails, this field would be null.
If this is not marked as nullable you would get an GraphQL error, which says that the field can't be nullable, and this error masks the actual error and does not return rejection message. So we should always heep these fields nullable.

vyshnavi1605 (approved):

martin-trajanovski (approved):

PR #565, opened 2022-02-16
fix(proposal permissions): Add instrument manager to queries like instrument scientist

bolmsten:
## Description

A instrument manager has not been able to see proposals on the instrument as the queries fetching proposals only searched for instrument scientists. This PR adds a where clause for the instrument manager as well. 

## Motivation and Context

To allow instrument managers to see proposals on a instrument

## Fixes

SWAP-2269

## Changes

Datasources for instruments, proposals and calls


src/datasources/postgres/InstrumentDataSource.ts
```diff
@@ -168,6 +168,8 @@ export default class PostgresInstrumentDataSource
         'i.instrument_id': 'ihs.instrument_id',
       })
       .where('ihs.user_id', userId)
+      .orWhere('i.manager_user_id', userId)
+      .distinct()
```
martin-trajanovski, line 172: Maybe it is better to use `distinct` with column name or `distinctOn` with column because not sure what really happens when you leave it empty https://knexjs.org/#Builder-distinct it is using `'*'` distinct on all columns. Maybe that is fine as well just trying to understand it better because it is harder like this without any column in the distinct.

src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -418,21 +418,23 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
...
+      .join('instruments', {
+        'instruments.instrument_id': 'proposal_table_view.instrument_id',
       })
+      .where(function () {
+        this.where('instrument_has_scientists.user_id', scientistId).orWhere(
+          'instruments.manager_user_id',
+          scientistId
+        );
+      })
+      .distinct()
```
martin-trajanovski, line 437: Same goes here I guess it works but it is hard to understand what column we want to distinct on

martin-trajanovski (approved):
Just small comment about the `distinct` method otherwise it looks good

PR #568, opened 2022-02-21
Optimise user officer proposal table

keiranjprice101:
## Description

Backend changes for UserOfficeProject/user-office-frontend#768

## Motivation and Context

Full motivation and context available in the linked related issue: UserOfficeProject/user-office-frontend#768

## How Has This Been Tested
Manual testing and test suites are passing.

## Changes
The backend changes for the previously linked issue. This moves the column sorting into the backend as well as the search. The search is performed using an `ILIKE` in the database. While I am not sure what the original search method is within material-table-core, the new method seems to perform fine during testing.

## Depends on

Backend changes for UserOfficeProject/user-office-frontend#768


src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -335,9 +341,85 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
...
+
+          query
+            .whereRaw('proposal_table_view.proposal_id ILIKE ?', searchText)
+            .orWhereRaw('proposal_table_view.title ILIKE ?', searchText)
+            .orWhereRaw('proposal_status_name ILIKE ?', searchText)
+            .orWhereRaw('instrument_name ILIKE ?', searchText);
+        }
+
+        if (sortField && sortDirection) {
+          switch (sortField) {
```
TickleThePanda, line 359: Instead of a switch, could we have a dictionary? Without types:
```ts
const fieldMap = {
    'proposalId': 'proposal_table_view.proposal_id',
    ...
}

const dbSortColumn = fieldMap[sortField];
```

There may be a better way to handle this but I'm not sure.

src/datasources/mockups/ProposalDataSource.ts
```diff
@@ -94,9 +94,14 @@ export class ProposalDataSourceMock implements ProposalDataSource {
...
   async getProposalsFromView(
-    filter?: ProposalsFilter | undefined
-  ): Promise<ProposalView[]> {
-    return [];
+    filter?: ProposalsFilter | undefined,
+    first?: number | undefined,
+    offset?: number | undefined,
+    sortField?: string | undefined,
+    sortDirection?: string | undefined,
+    searchText?: string | undefined
```
jekabs-karklins, line 102: I wonder why is there a new field `searchText`.
Could not you reuse the `ProposalsFilter.text` from the `filter` parameter

jekabs-karklins (approved):
Otherwise it looks good from my side 💯

src/datasources/postgres/ProposalDataSource.ts
keiranjprice101: Yes the switch statement grew quite large as it was built column by column. This might work nicely, though not every column has to be mapped like title. So this means either the dictionary will have equal key,value pairs or a try catch. But I think this is a good suggestion cheers 👍

src/datasources/mockups/ProposalDataSource.ts
keiranjprice101: Thanks, I will give this a try. I am not familiar with the filters at all really, so it didn't occur to me. 👍

keiranjprice101: After taking a look  I see that using the text filter will apply 2 separate ILIKES on some of the same fields. Introducing searchText means only `ILIKE` will ever be applied for the user office table, and any other part of the frontend using the query will remain unaffected. So I think it makes more sense to keep it as is, unless you have some other thought?

src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -335,9 +355,41 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
...
+
+          query
+            .whereRaw('proposal_table_view.proposal_id ILIKE ?', searchText)
+            .orWhereRaw('proposal_table_view.title ILIKE ?', searchText)
+            .orWhereRaw('proposal_status_name ILIKE ?', searchText)
+            .orWhereRaw('instrument_name ILIKE ?', searchText);
+        }
+
+        if (sortField && sortDirection) {
+          sortField = fieldMap[sortField];
```
TickleThePanda, line 373: What happens if the `sortField` is not found here? I think we should throw a pre-emptive exception.

keiranjprice101: I am not sure how that could happen unless querying directly. Do you mean to check the field and to throw a different exception before putting it into the map?

TickleThePanda: As this is also going to be a developer facing API, querying the GraphQL API directly is possible.

If the `fieldMap[sortField]` is the "not found" value (I don't know if it's `null` or `undefined`) then we should throw an exception.

PR #570, opened 2022-02-24
feat: Add visitStart and visitEnd dates

jekabs-karklins:
## Description

This PR aims to add start and end visit dates to the database. This information, later on, will be shown on the experiments page

## Motivation and Context

In the Experiments Page, we will need to show visits along with their start and end dates.

## How Has This Been Tested

Added and updated E2E

## Changes

- new DB patch
- new fields

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/775


src/models/questionTypes/FileUpload.ts
```diff
@@ -1,5 +1,5 @@
-/* eslint-disable quotes */
 import { fileUploadQuestionValidationSchema } from '@user-office-software/duo-validation';
+/* eslint-disable quotes */
```
martin-trajanovski, line 2: Is this still needed or? Not sure why it is moved one line bellow.

martin-trajanovski (approved):
Looks great!

src/models/questionTypes/FileUpload.ts
jekabs-karklins: It's needed because of 
`return queryBuilder.andWhereRaw(
            "jsonb_array_length((answers.answer->>'value')::jsonb) > 0"
          );
`
It tries to replace double quotes with single quotes and nested singlequotes it tries to escape, but that breaks the SQL query

PR #576, opened 2022-03-01
feat: Show experiment times

jekabs-karklins:
## Description

This PR aims to implement a feature where the user offer has an overview of what are the upcoming experiments. 
In this view, officers can see also visit information, and also proposal ESI.

This view could be extended to add more relevant information for user officers to see other information about upcoming experiments

Peek 2022-03-03 14-04

## Motivation and Context

Currently, there was no way for the user officer to see the list of upcoming experiments or visits

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1635

## Changes

ScheduledEventDataSource new endpoints

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/787


src/resolvers/types/ScheduledEvent.ts
```diff
@@ -32,10 +32,10 @@ export class ScheduledEventCore {
   @Field(() => ScheduledEventBookingType)
   bookingType: ScheduledEventBookingType;
 
-  @Field(() => TzLessDateTime)
+  @Field(() => Date)
```
martin-trajanovski, line 35: I guess this was changed on purpose because previous type was formatting the date and removing the time zone I think. Now everything is sent.

martin-trajanovski (approved):
Looks good!

keiranjprice101 (approved):
I have taken a quick look over this, as I was automatically assigned. But I do not believe we use the visits functionality at the STFC? And I am not familiar with it, but I can take a more thorough look if necessary

src/resolvers/types/ScheduledEvent.ts
jekabs-karklins: I changed for consistency, because in most places we use Date. But maybe we should change to use in all places TzLessDateTime. We should talk about it.

PR #579, opened 2022-03-04
fix: add more logs to verify user error

jekabs-karklins:
## Description

Currently, there is incomplete information about the error when it happens. This PR will add more details to the 'can not verify user' error log


martin-trajanovski (commented):
Not sure if we are exposing too much information like this adding `decoded` object to the logs or it is fine.

bolmsten (approved):

PR #583, opened 2022-03-07
fix: Small possibility of duplicated proposal_ids (UserOfficeProject/user-office-project-issue-tracker#299)

RussellMcLean:
## Description

This PR avoids the creation of duplicate proposal_id when a new proposal is submitted and there are conflicting proposal_ids already in the database (e.g. imported legacy data). It does NOT account for the case where a new legacy data import conflicts with a proposal_id already submitted.

## Motivation and Context

When a new proposal is submitted, the generated id may conflict with an id already in the database. This would throw an error with message 'duplicate key value violates unique constraint "proposals_short_code_key"'. The unique constraint is enforced from the [.sql script](https://github.com/UserOfficeProject/user-office-backend/blob/develop/db_patches/0002_AlterProposals.sql?rgh-link-date=2021-12-03T11%3A55%3A19Z) which generates the proposal_id. Inserting a new proposal into the database would fail as a result of this error.

## How Has This Been Tested

- manual tests.

## Fixes

Closes [#299](https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/299)

## Changes

- added a db patch [0118_AlterGenerateProposalShortcode.sql](https://github.com/UserOfficeProject/user-office-backend/blob/stfc-user-office-project-299-duplicated-proposal-ids/db_patches/0118_AlterGenerateProposalShortcode.sql) which will retry generating the proposal_id until there is no conflict. Note that race conditions should not be a problem during proposal_id generation since the function is locked with the PARALLEL UNSAFE argumant by default.


simonfernandes (commented):
It looks like this will fix the issue and work well, but I wonder if a slightly better place for this logic (or similar) would be in the Postgres function that generates the proposal_id itself. It would have the advantage of being more efficient (only a single request ever made), not limited to 5 retries, and keeps proposal_id generation tied to the same place.

This is the `generate_proposal_shortcode` function which you can browse to and view in [pgAdmin](https://user-office-project.atlassian.net/wiki/spaces/UOP/pages/19136712/STFC+development+workflow#Using-postgres), which is run after the insert has been made in the data source.

It looks like it's possible to use selects and loops in Postgres functions, so it might be possible to do something like this in there:

```
query = select from proposals where proposal_id = generated_id
while query returns a row:
    generate another id
```

RussellMcLean:
Something like this should be possible and would be the more logical place for the fix. But would there be any significant impact to performance if it has to search the list of proposal_ids for every new proposal?

TCMeldrum (approved):

jekabs-karklins (approved):
Looks good

PR #584, opened 2022-03-08
fix: boolean question filter not working

jekabs-karklins:
## Description

This PR aims to fix the issue where boolean question filter not working

## Motivation and Context

The answer to the boolean question is considered as NO if the answer is `FALSE` or the answer does not exist.
This PR adds an argument `isNot` to the GRAPHQL endpoint when searching for proposals by question answer. If `isNot` is set to TRUE, the query will invert the result.

For the specific boolean question problem, in order to get all proposals where the question was answered with `FALSE` or not answered at all, we ask for proposals in which the boolean question is not "TRUE"

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2350

## Changes

Added new parameter
Updated SQL query

## Depends on
https://github.com/UserOfficeProject/user-office-frontend/pull/801


martin-trajanovski (approved):
Looks good! I hope the new seed wont make any difference in the e2e tests where we call resetDB(true).

PR #586, opened 2022-03-08
feat: validate file upload types

simonfernandes:
## Description

Adds validation for file uploads by checking the file signature of the file and ensuring that its type is one of the accepted file types for the question. (See changes below for more info)

Caveats with this method:
- Files may not be able to be identified (we agreed to reject in this case)
- There is some overlap in file signatures (e.g. an excel spreadsheet, word document and zip file may all have the same signature). 

An npm library called magic-bytes.js was used here for the signature checking. We had previously spoken about a more popular library called file-type, but the developer has made the divisive decision to make it a ES module only with no support for CommonJS.

## Motivation and Context

We would like to restrict users to only upload a certain type of file

## How Has This Been Tested

- Manual testing
- Created e2e tests

## Fixes

https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/226

### Frontend validation
- Check that the file about to be uploaded has a file extension, don't allow upload if not
- Check the request has the correct Content-Type header, don't allow upload if not

### Backend validation

(A file upload question can consist of multiple files)

Validation for an individual file passes if:
- The file can be read and is determined to be one of the accepted types
- If there is a problem retrieving the file from the database (i.e. a problem on our end)

Validation for an individual file fails if:
- The file can is determined to be of a different type
- The file cannot be determined to be one of the required types (i.e. the file is not recognised)

Validation for the question passes if all individual file checks (above) pass, and fails if one or more of the individual file validation fails.

### Other
- Require a file type to be chosen when creating a file upload question
- Accepted file type information is displayed alongside the question

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/803


src/models/questionTypes/FileUpload.ts
```diff
@@ -1,20 +1,95 @@
...
+import { isRejection } from '../Rejection';
 import { DataType } from '../Template';
 import { Question } from './QuestionRegistry';
 
 export const fileUploadDefinition: Question = {
   dataType: DataType.FILE_UPLOAD,
-  validate: (field: QuestionTemplateRelation, value: any) => {
-    if (field.question.dataType !== DataType.FILE_UPLOAD) {
-      throw new Error('DataType should be FILE_UPLOAD');
+  validate: async (field: QuestionTemplateRelation, value: any) => {
```
jekabs-karklins, line 19: Maybe we can move the validation into a separate file?

jekabs-karklins (approved):

src/models/questionTypes/FileUpload.ts
simonfernandes: I think that's a good idea, is there a suitable place in the backend for it? I wasn't sure where it would go. I thought it would go in user-office-validation package but that seems to be read only now.

jekabs-karklins: You are right, about the suitable place. Maybe it is sufficient to factor it out into separate function within the same file.

PS: `user-office-validation` has been moved into the `user-office-lib` repository but that would not be the best place.

src/models/questionTypes/FileUpload.ts
```diff
@@ -1,20 +1,35 @@
...
-    return fileUploadQuestionValidationSchema(field).isValidSync(value);
+    const typeCheckResults: boolean[] = value.map(
+      async (file: { id: string }) => {
+        return await isValidFileType(file.id, field, value);
+      }
+    );
+
+    return (await Promise.all(typeCheckResults)).every((check) => {
+      return check;
+    });
```
ACLay, line 32: What does this block do? It seems like the promise wrapping doesn't do anything, and would be equivalent to `return typeCheckResults;`

src/models/questionTypes/FileUpload.ts
```diff
@@ -48,3 +63,79 @@ export const fileUploadDefinition: Question = {
...
+
+  const path = await fileMutations.prepare(fileId);
+
+  if (isRejection(path)) {
+    logger.logInfo(
+      'Could not retrieve file to validate file upload question',
+      errorContext
+    );
+
+    return true; // Allow the file
```
ACLay, line 99: The method's docstring says "returns... False if the file cannot be retrieved", but here we log that it can't be retrieved and then return true. Same on line 112

ACLay (changes requested):

src/models/questionTypes/FileUpload.ts
simonfernandes: The `promise.all` wrapping waits for each promise to resolve, which is necessary because when using `array.every()`, the value of a pending promise is equal to true. It's also like this because `array.every()` doesn't work with async functions

ACLay:
So to check I'm understanding this, returning true and the validation passing if we can't find the file is desired behaviour. Is this is to allow optional file upload questions? Presumably then if a file is required, this would be used in conjunction with another validator to check an answer was provided?

ACLay (approved):

PR #596, opened 2022-03-11
feat: updating shipment label

jekabs-karklins:
## Description

This PR aims to update the label to what we discussed in the latest EAM meeting.
The following changes have been made:
- Added local contact
- Updated facility logo
- Updated additional details
- Using new units

## Motivation and Context
We want to have a label similar to the end result so that we can show stakeholders how is it going to look

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2360

## Depends on

https://github.com/UserOfficeProject/user-office-factory/pull/93


martin-trajanovski (approved):
Looks good!

PR #597, opened 2022-03-14
refactor: Date and Datetime format standardization and use of luxon for datetime manipulation

martin-trajanovski:
## Description

This PR aims to standardize the date and datetime formats used in the user office software. Previously we were modifying the dates a lot with different kind of formats throughout the app.
- Remove custom `TzLessDateTime` scalar type, use `luxon` instead of moment for dates
- Now all the formats should be kept as setting in the database and use for rendering dates on the frontend in the same format.
- This also fixes some issues with date search filter when exact date is used to filter by.

## Motivation and Context

Have one standard for rendering dates and keep that at one place for easier maintenance and easier modifications based on facilities.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2345

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/815


src/models/questionTypes/FileUpload.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable quotes */
```
keiranjprice101, line 1: Is there a reason it's easier to add this disable instead of swapping the 2 instances of double quotes in this file?

db_patches/0118_AddDateTimeFormatSetting.sql
```diff
@@ -0,0 +1,18 @@
...
+	IF register_patch('AddDateTimeFormatSetting', 'Martin Trajanovski', 'Add date time format setting', '2022-03-14') THEN
+		BEGIN
+		  INSERT INTO 
+			settings(settings_id, description)
+		  VALUES
+			('DATE_FORMAT', 'Format used to represent date without times');
+		  INSERT INTO 
+			settings(settings_id, description)
+		  VALUES
+			('DATE_TIME_FORMAT', 'Format used to represent date with time whouut seconds.');
```
keiranjprice101, line 13: minor typo here `whouut`

keiranjprice101 (approved):
Looks good to me, 2 minor questions

src/models/questionTypes/Date.ts
```diff
@@ -28,9 +29,19 @@ export const dateDefinition: Question = {
     (relation.config as DateConfig).defaultDate || null,
   filterQuery: (queryBuilder, filter) => {
     const value = JSON.parse(filter.value).value;
+    // NOTE: For exact comparison we need to use like interval of whole day because for now filter works only with date without time.
```
jekabs-karklins, line 32: I'm not sure if we need to compare the whole day, because for date inputs the time component should be 00:00:000.

Did you have any issues with this? Maybe I can take a look?

martin-trajanovski: Yes I had issues when time is included in the question answer I think. You can doublecheck this but try that scenario (Include time and try to search for exact date).

src/models/questionTypes/FileUpload.ts
martin-trajanovski: I think all the question types have this because of the queries and maybe @jekabs-karklins can help more with that.

src/models/questionTypes/Date.ts
jekabs-karklins: OK, I believe this is actually correct 👍🏻

jekabs-karklins (approved):

PR #617, opened 2022-04-01
Speed up getting templates in STFC mode

TCMeldrum:
Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/518

## Description

When getting templates it is very slow in safe mode this is because we are making a lot of calls to the UOWS to get users' roles. 

Done a similar refactor to [this ](https://github.com/UserOfficeProject/user-office-backend/pull/543)to cut down on the number of calls to UOWS. Made this fix while working on another issue but it should be its own issue.

This change broke most of the tests as they no longer had permissions, I have now coded them into the dummy users. I don't think this invalidates any of the tests but I am not entirely sure. 


simonfernandes (approved):

jekabs-karklins:
> @UserOfficeProject/swap Had a chat with @TickleThePanda and we were a little confused what is the purpose of the `hasRole` check. I believe the user role that we are checking comes from the JWT token and there can be trusted?

@TCMeldrum when I read this I think you are right that the `hasRole` has been moved into JWT token, and an be considered as trusted. From what I can understand reading the code of the slow part that has been removed it's purpose was to check if the roles of users has not changed since JWT token was issued.
Therefore I think storing hasRole value in the JWT token servers no purpose and the slow par could be removed if we want to accept that user can still perform some role actions after the role has been removed until JWT token expires and new one is obtained.

martin-trajanovski:
> > @UserOfficeProject/swap Had a chat with @TickleThePanda and we were a little confused what is the purpose of the `hasRole` check. I believe the user role that we are checking comes from the JWT token and there can be trusted?
> 
> @TCMeldrum when I read this I think you are right that the `hasRole` has been moved into JWT token, and an be considered as trusted. From what I can understand reading the code of the slow part that has been removed it's purpose was to check if the roles of users has not changed since JWT token was issued. Therefore I think storing hasRole value in the JWT token serves no purpose and the slow par could simply be removed if we want to accept that user can still perform some role actions after the role has been removed until JWT token expires and new one is obtained.

@TCMeldrum @jekabs-karklins I think this was added to check anyway even if there is a currentRole inside the token because you could remove that role from a user and it will be still usable for some time if we remove that database check. But if you want to accept that for some amount of time maybe on user role removal we should mention that as a warning that if the user uses that as a current role at the moment he/she will be able to perform some actions for some amount of time. Maybe creating a task for this would be a good idea. Just to have the warning in the frontend so the one who removes the role knows what to expect.

TickleThePanda:
> If you want to accept that for some amount of time maybe on user role removal we should mention that as a warning that if the user uses that as a current role at the moment he/she will be able to perform some actions for some amount of time.

I'm happy with this - I don't think there's a huge risk involved in that. A warning in the UI sounds reasonable.

TCMeldrum:
I have removed the check and opened a PR for the frontend to add a warning: https://github.com/UserOfficeProject/user-office-frontend/pull/863

TickleThePanda (approved):

PR #621, opened 2022-04-05
fix: review must have grade when submitting

jekabs-karklins:
## Description

Preventing submitting a review without grade

## Motivation and Context

Reviewer should not be able to submit review without grade and comment

## Fixes

https://jira.esss.lu.se/browse/SWAP-2477

## Changes

Added validation schema
Added business logic

## Depends on
https://github.com/UserOfficeProject/user-office-frontend/pull/847

### DIFF
--- a/src/mutations/ReviewMutations.ts
+++ b/src/mutations/ReviewMutations.ts
@@ -149,7 +149,8 @@ export default class ReviewMutations {
       );
     }
 
-    if (!review.grade || !review.comment) {
+    const isReviewValid = await proposalGradeValidationSchema.isValid(review);
+    if (isReviewValid === false) {
       return rejection(
         'can not submit proposal review because grade or comment is missing',
         { args }
--- a/src/mutations/ReviewMutations.ts
+++ b/src/mutations/ReviewMutations.ts
@@ -2,6 +2,7 @@ import {
   proposalGradeValidationSchema,
   proposalTechnicalReviewValidationSchema,
   addUserForReviewValidationSchema,
+  submitProposalReviewValidationSchema,
 } from '@user-office-software/duo-validation';
 import { container, inject, injectable } from 'tsyringe';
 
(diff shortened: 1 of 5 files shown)


src/mutations/ReviewMutations.ts
```diff
@@ -148,6 +149,13 @@ export default class ReviewMutations {
       );
     }
 
+    if (!review.grade || !review.comment) {
```
martin-trajanovski, line 152: I think this isn't necessary because validation should check it as grade and comment are required.

martin-trajanovski: Ah ok I see now that you changed the validation schema and the input you send here. Maybe it is still needed in the end.

src/resolvers/mutations/SubmitProposalsReviewMutation.ts
```diff
@@ -36,18 +36,22 @@ export class SubmitProposalsReviewMutation {
...
-    const results = await Promise.all(
-      submitProposalsReviewInput.proposals.map((proposal) => {
-        return context.mutations.review.submitProposalReview(
-          context.user,
-          proposal
-        );
-      })
-    );
+    const failedProposals = [];
+    for await (const proposal of submitProposalsReviewInput.proposals) {
```
martin-trajanovski, line 40: Is this way more efficient from the previous `Promise.all` because I think Promise.all sends all requests at once and wait for the last one to arrive. And the current approach that you have now sends them one by one.

src/mutations/ReviewMutations.ts
jekabs-karklins: Yes, kind of, since grade and comment is not send in the submit request.

src/resolvers/mutations/SubmitProposalsReviewMutation.ts
jekabs-karklins: This is changed it for the readability/debuggability sake

src/mutations/ReviewMutations.ts
jekabs-karklins: I think that would be even better, yes. Then we keep validation in one place. I will look into that.

martin-trajanovski: Great. I am not touching this part anyway on the backend so would be fine for you to just fix and I will update my frontend part using your way of validation as well. 🚀

martin-trajanovski (approved):

PR #632, opened 2022-04-21
Sub templates imports and scripts to import templates 

TCMeldrum:
Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/338

## Description

This PR adds the ability to import subtemplates for generic templates and sample questions on templates. It added the subtemplates as strings in an array onto the JSON. Then when importing the template it will first import the subtemplates make a note of the new template ID then import the rest of the template. There might still be some fuzziness with the template Ids but I think it is working. 

There were a few other bug fixes. You could not delete some curtain proposal/ generic templates as they were in the active templates table despite that table not being relevant to proposal/generic templates. Generic template questionarys not being deleted their corresponding proposal questionary were deleted. API keys could not delete proposals.

## Motivation and Context

Allows to fully import templates and make our dev environment mirror our prod environment. 

## Changes

- Adds importing of subtemplates on proposal templates
- Fixes generic template questionaries not being deleted
- Fixes some templates being undeletable
- API keys can now delete proposals

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/871


src/auth/UserAuthorization.ts
```diff
@@ -35,6 +35,14 @@ export class UserAuthorization {
     return agent?.currentRole?.shortCode === Roles.USER;
   }
 
+  isApiToken(agent: UserWithRole | null) {
+    if (agent == null) {
```
martin-trajanovski, line 39: I guess we can use triple equal here `agent === null`

martin-trajanovski (approved):
Just a small comment otherwise it looks good.

simonfernandes (approved):

PR #634, opened 2022-04-25
feat: send Rapid Access confirmation emails to the STFC User Office 

simonfernandes:
## Description

This adds a few STFC emails and makes some adjustments to the email handler to send them.

Because STFC now wants to send different emails, I thought this was the right time to refactor the email handler into separate STFC and ESS handlers. The (new) email handler for the ESS has everything copied from the old one. I did make sure to rename the file in Git but I think the diff is too big so it detects it as a new file, sorry about that.

## Motivation and Context

We need to send our User Office an email when certain proposals are submitted. We also need to send users different emails based on the facility they are submitting to.

## How Has This Been Tested

For STFC, I have created different calls and submitted proposals to them, checking that the correct emails are received. I have also checked the email content with Emma Gozzard.

For ESS, I have used the e2e config (using the ESS handler) and submitted a proposal. The logs show an email was sent. 

## Fixes

Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/506 and https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/524.

Also fixes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/286

## Changes

- Creates separate handlers for the ESS and STFC
- Sends new STFC emails. For now this is based on the title of the call until we have the concept of facilities.
  - Sends an email to the User Office when a Rapid proposal is submitted
  - Sends a Rapid email to the PI when a Rapid proposal is submitted
  - Sends an ISIS email to the PI when an ISIS proposal (except Rapid) is submitted. It's slightly different to the Rapid email and I thought it may as well be included as part of this issue.
- Fixes an existing issue where CSS wasn't being applied to emails

## Depends on

docker-orchestration: https://github.com/isisbusapps/docker-orchestration/pull/246
FBS branch: `user-office-project-506-rapid-email-notifications`


src/eventHandlers/email/essEmailHandler.ts
```diff
@@ -0,0 +1,248 @@
...
+          },
+          substitution_data: {
+            title: 'ESS User reset account password',
+            buttonText: 'Click to reset',
+            link: event.userlinkresponse.link,
+          },
+          recipients: [{ address: event.userlinkresponse.user.email }],
+        })
+        .then((res: any) => {
+          logger.logInfo('Emai send on for password reset:', {
```
keiranjprice101, line 38: Emai

mr-field (commented):
The logic looks good, but do we really need to split the email handlers? I think we could keep everything in the same handler and just change the template or send the additional email based on the call's facility.

About the email templates in BisAppSettings, I noticed a couple of things:
- The ISIS direct email (`isis-proposal-submitted-pi`) seems identical to the one for rapid (`isis-rapid-proposal-submitted-pi`), except for a line telling people they can edit their proposals in Rapid. Is there really any difference in the new system? I though they would all be editable after submission. We should use the same template if so
    ```
    < <p>You or any of the experimenters listed above may view this proposal via our <a href="https://proposal.facilities.rl.ac.uk/">proposal submission site</a>.</p>
    ---
    > <p>You or any of the experimenters listed above may view or edit this proposal. Visit our <a href="https://proposal.facilities.rl.ac.uk/">proposal submission site</a> to make changes to this proposal.</p>
    ```
- A few of the templates have this line, which has a duplicated `</p>` tag at the end. We should remove the duplicate:
   ```
   <p>If you wish to withdraw your involvement with this proposal or you wish your name to be removed from the database, please contact the <a href="mailto:isisuo@stfc.ac.uk">ISIS User Office</a>.</p></p>
   ```

simonfernandes:
> The logic looks good, but do we really need to split the email handlers? I think we could keep everything in the same handler and just change the template or send the additional email based on the call's facility.

I don't think it was strictly necessary as you say, but my thinking was that adding STFC specific logic in the shared handler felt messy and potentially confusing. I was under the impression that splitting it was something that we always wanted to do, and it was only acceptable now while our requirements roughly align. If you think the single handler would be better, I can change it back.

> The ISIS direct email (isis-proposal-submitted-pi) seems identical to the one for rapid (isis-rapid-proposal-submitted-pi), except for a line telling people they can edit their proposals in Rapid. Is there really any difference in the new system? I though they would all be editable after submission.

The workflow that the call uses determines whether the proposal is editable after submission. Having checked with Emma, we don't want Rapid proposals to be editable, and the email should have that line removed. I could have used a conditional in a single template, but we don't test the templates so I thought it best that logic was kept out of them.

simonfernandes:
I've now removed those duplicated tags from the emails in the FBS branch

mr-field:
> I don't think it was strictly necessary as you say, but my thinking was that adding STFC specific logic in the shared handler felt messy and potentially confusing. I was under the impression that splitting it was something that we always wanted to do, and it was only acceptable now while our requirements roughly align.

My personal preference would be to avoid splitting the app into facility-specific parts unless we really have to. This is mostly because once a component is split, each facility is responsible for their own implementation and they won't pick up any features or fixes added by another facility. We never really discussed this, so maybe it's best to hear what the @UserOfficeProject/swap team thinks.

The email changes look good to me given that we really don't want to edit Rapid proposals after submission.

simonfernandes:
That's fair enough, and thanks for the reasoning. I'll hold off then until we get others' thoughts.

I think I was recalling this issue: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/106. Admittedly I created it, though I seemed to remember that in sprint planning we said it was something we'd like to do and sized it. That was a long time ago now, though.

bolmsten (approved):

mr-field (approved):

PR #635, opened 2022-04-26
feat: Remove access for SEP chair and secretary to un-submit or change submitted reviews

martin-trajanovski:
## Description

SEP Chair and SEP Secretary should not be able to change already submitted reviews and can not un-submit as well.

## Motivation and Context

Previously SEP Chair and Secretary were allowed to do any action on the SEP reviews.

## How Has This Been Tested

- manual tests
- unit tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2474

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/873

### DIFF
--- a/src/auth/ReviewAuthorization.ts
+++ b/src/auth/ReviewAuthorization.ts
@@ -83,22 +83,16 @@ export class ReviewAuthorization {
   ): Promise<boolean>;
   async hasWriteRights(
     agent: UserWithRole | null,
-    reviewId: number,
-    reviewAlreadySubmitted?: boolean
-  ): Promise<boolean>;
-  async hasWriteRights(
-    agent: UserWithRole | null,
-    reviewOrReviewId: Review | number,
+    review: Review,
     reviewAlreadySubmitted = false
   ): Promise<boolean> {
-    const review = await this.resolveReview(reviewOrReviewId);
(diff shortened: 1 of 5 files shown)


src/auth/ReviewAuthorization.ts
```diff
@@ -83,24 +83,18 @@ export class ReviewAuthorization {
...
   async hasWriteRights(
     agent: UserWithRole | null,
-    reviewId: number,
-    reviewAlreadySubmitted?: boolean
-  ): Promise<boolean>;
-  async hasWriteRights(
-    agent: UserWithRole | null,
-    reviewOrReviewId: Review | number,
+    review: Review,
     reviewAlreadySubmitted = false
```
jekabs-karklins, line 87: Just a small question here, why is `reviewAlreadySubmitted` sent as a parameter, can that be looked up by the authorizer itself?

jekabs-karklins (approved):

src/auth/ReviewAuthorization.ts
martin-trajanovski: Yes I think I sent it like this before but now I changed some logic and you are right this could be improved.

PR #637, opened 2022-04-26
fix: Fixing possible race condition when calling resetDB

martin-trajanovski:
## Description

When getting or trying to mutate some data right after the `resetDB` call we were getting errors all the time because call finishes before setting up the environment properly. This was because we were not waiting for environment configure to finish.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2444

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/875


jekabs-karklins (approved):

jekabs-karklins:
Nice one 💯

simonfernandes (approved):

PR #644, opened 2022-04-28
fix: EAM expects SI values

jekabs-karklins:
## Description

EAM is expecting values in SI base units (meters, kilograms)

## Motivation and Context

We want proper units to be recorded in the EAM system

## How Has This Been Tested

- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2526

## Depends on

N/A


martin-trajanovski (approved):
Looks good!

PR #647, opened 2022-04-29
feat: Run an antivirus scan on uploaded files

ACLay:
## Description

Adds an antivirus scan check to file upload question validation using the [Clamscan](https://www.npmjs.com/package/clamscan) package to interface with an external Clam antivirus service over TCP.

If environment variables `ANTIVIRUS_HOST` and `ANTIVIRUS_PORT` are not both specified the check will be skipped, and if the service cannot be reached the check will also pass. The aim of this was to mirror the current behaviour of the file type check and to not prevent users being able to submit if the service is unavailable for innocent network issues. However, if the preference of reviewers is to have the check fail if the service can't be reached I'm happy to make that change.

The clam antivirus service can be provided in a separate container with simple docker compose config:
```yaml
  clam-antivirus:
    image: clamav/clamav:0.104
    ports:
      - "3310:3310"
```
then providing `clam-antivirus` and `3310` to the backend `ANTIVIRUS_*` variables. This is done in https://github.com/isisbusapps/docker-orchestration/pull/248 for the STFC docker config

## How Has This Been Tested

I've tested the integration by running the new code with and without environment variables, and with and without the clam antivirus container running. In all instances I've been able to pass the file upload validation, and when my files are sent for scanning, they get passing responses.

I have not tested this with any sort of malicious file.

## Fixes

Closes userofficeproject/user-office-project-issue-tracker#327

### DIFF
--- a/src/models/questionTypes/FileUpload.ts
+++ b/src/models/questionTypes/FileUpload.ts
@@ -100,9 +100,9 @@ export const fileUploadDefinition: Question = {
  * Checks the file signature of a file and returns whether the file
  * is one of the accepted types specified in the question config.
  *
- * @param fileId The ID of the file to check
+ * @param path A valid path to the file to be scanned
  * @param field The question data containing the config
- * @param value All files in the question
+ * @param errorContext File details to be logged in the event of an error
  * @returns True if the file is identified as one of the accepted
  * types, or if the file cannot be retrieved. False is the file
  * cannot be identified, or is not one of the accepted types.
--- a/src/models/questionTypes/FileUpload.ts
+++ b/src/models/questionTypes/FileUpload.ts
@@ -3,6 +3,7 @@ import fs from 'fs/promises';
 
 import { logger } from '@user-office-software/duo-logger';
 import { fileUploadQuestionValidationSchema } from '@user-office-software/duo-validation';
+import NodeClam from 'clamscan';
 import fileTypeInfo from 'magic-bytes.js';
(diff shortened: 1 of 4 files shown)


src/models/questionTypes/FileUpload.ts
```diff
@@ -79,29 +108,10 @@ export const fileUploadDefinition: Question = {
  * cannot be identified, or is not one of the accepted types.
  */
 const isValidFileType = async (
-  fileId: string,
+  path: string,
```
simonfernandes, line 111: The docstring needs updating with the changed arguments

src/models/questionTypes/FileUpload.ts
```diff
@@ -142,3 +150,53 @@ const isValidFileType = async (
...
+    .then((clamAV) => clamAV.isInfected(path))
+    .then((response) => {
+      const isInfected = response.isInfected;
+      if (isInfected) {
+        logger.logError('Infected file was uploaded', {
+          response,
+          errorContext,
+        });
+      } else {
+        logger.logInfo('Virus scan complete', { response });
```
simonfernandes, line 189: It looks like `isInfected` can return null if the file is unable to be scanned, so I don't think this would strictly be true. It wouldn't be thrown, caught and logged below either. Even though the response in this log would contain the null value, maybe we should separately log in this eventuality

jekabs-karklins (approved):
💯

PR #653, opened 2022-05-05
#234 Add linter workflow

charlieDurnsfordHollands:
## Description
Part of: UserOfficeProject/stfc-user-office-project/issues/234

## Motivation and Context
Adding linter to PRs to maintain consistency and stop any issues not spotted locally

## How Has This Been Tested
Tested on this PR branch 

## Fixes
Closes: UserOfficeProject/stfc-user-office-project/issues/234 

## Changes

Added linter .yml file that runs on pull request and gives prettier and eslint errors


martin-trajanovski:
@charlieDurnsfordHollands I think we can remove the `dev-commands` folder because nobody uses that anymore and it is not maintained. The idea there was to have a command where we create some CRUD operations for new endpoint but since then it our process changed a bit and this is not updated. @jekabs-karklins what do you think about this removing of dev-commands and everything related to it? It will probably make the linting process a bit faster because those files are plain javascript.

TickleThePanda:
> The Prettier errors point out which files the errors are in but not the specific files, I believe getting a more detailed error would involve needing permissions, is this something that we want?
> 
> "The action doesn't have permission to create annotations for commits on forks unless you use the pull_request_target event. You can modify the default permissions granted to the GITHUB_TOKEN by using the permissions key and set the checks scope to write. See [GitHub documentation](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions) for more."

I'm not sure that this message means that. I think this is talking about annotations for [GitHub Forks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/about-forks).

I don't think there's an easy way to get prettier to give the details as it's not a linter, it's a formatter - so I think it's okay for it to say "this isn't formatted correctly, please run the formatter.

---

Also, I can see this for the backend but I think we should have an equivalent for the frontend and factory.

jekabs-karklins:
> @charlieDurnsfordHollands I think we can remove the `dev-commands` folder because nobody uses that anymore and it is not maintained. The idea there was to have a command where we create some CRUD operations for new endpoint but since then it our process changed a bit and this is not updated. @jekabs-karklins what do you think about this removing of dev-commands and everything related to it? It will probably make the linting process a bit faster because those files are plain javascript.

@martin-trajanovski it's fine by me to remove it

martin-trajanovski (approved):

TickleThePanda (changes requested):
You won't be able to merge this until you've either (chosen appropraitely) fix the style in each of the files or ignore the files.

RussellMcLean:
.github/workflows/* -> ignored
README.md -> ignored
api_tests/* -> formatting fix (indentation or clrf)
src/datasources/stfc/UOWSSoapInterface.ts -> formatting fix (indentation)

TickleThePanda (approved):

jekabs-karklins (approved):

PR #654, opened 2022-05-06
feat: Add ability to un-impersonate user after it gets impersonated

martin-trajanovski:
## Description

Add `impersonatingUserId` inside the jwt token to be able to get the information later when you want to un-impersonate and compare if it is correct one because now the endpoint is open for any authorized user.

## Motivation and Context

Previously we were able to only impersonate user as user officer but not go back with one click.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2392

## Changes

https://user-images.githubusercontent.com/6333063/168036262-bbcfcf6b-fd18-4461-b8fc-6abb7d5dfc84.mp4

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/895


src/models/User.ts
```diff
@@ -52,6 +53,7 @@ export class User {
 
 export interface UserWithRole extends User {
   currentRole: Role | undefined;
+  impersonatingUserId?: number | null;
```
jekabs-karklins, line 56: maybe it should be just 
`impersonatingUserId: number | undefined;`
same as `currentRole`

src/mutations/UserMutations.ts
```diff
@@ -408,11 +408,39 @@ export default class UserMutations {
...
+
+    // NOTE: This is checking if person trying to impersonate user is not user officer then reject.
+    if (isImpersonatingUser && !isUserOfficer) {
+      return rejection(
+        'Can not impersonate user because of insufficient permissions',
+        { userId }
+      );
+    }
+
+    if (!isImpersonatingUser && isUserOfficer) {
```
jekabs-karklins, line 429: The message says 'Can not impersonate user with already impersonated user'

Should not the if therefore be `if (isImpersonatingUser && isUserOfficer)` ? or am I missing something

simonfernandes:
I'm sorry @jekabs-karklins, I was reviewing at the same time and left a very similar comment to yours, but accidentally deleted yours instead of mine 🙁

src/mutations/UserMutations.ts
simonfernandes, line 429: Comment left by Jekabs which I accidentally deleted:

>The message says 'Can not impersonate user with already impersonated user'
>
>Should not the if therefore be if (isImpersonatingUser && isUserOfficer) ? or am I missing something

simonfernandes: I was also thinking the same thing, that the boolean would make more sense inverted. 

Could this check also be shortened to just `if (isImpersonatingUser)`, and the same with the user officer check: `if (!isUserOfficer)`?

simonfernandes (approved):

PR #656, opened 2022-05-11
refactor(auth): Moving UserAuth into dependencies

jekabs-karklins:
## Description

Previously UserAuth was resolved from the container by sending in UserAuthorization class. This effectively meant that new instance of the class was created and constructed attributed decorated with the @inject tag was resolved.

I have now refactored so that UserAuth is resolved by Token. This allows mapping UserAuthorization to do different implementations based on facility, or in other words, based on what is specified in the dependency configs.

This refactor is not mapping the UserAuthorization to facility-specific class yet in this PR. I wanted to keep PRs smaller so this refactor comes in first.

## Motivation and Context
We want to be able to interface out UserAuthorization

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

N/A

## Changes

UserAuthorization

## Depends on

N/A

N/A


martin-trajanovski (approved):
Looks good!

PR #657, opened 2022-05-12
feat(auth): As a user I want single sign on with PING

jekabs-karklins:
## Description

This PR aims to integrate ESS UserOffice with single-sign-on using the identity access management system PING, developed by our external partners.

## Motivation and Context
The approach for integrating with the external authorization system piggybacks on the external authorization developed by STFC and uses the same approach.

The difference is in UserAuthorization class, which both facilities implement differently.

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2564

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/914


martin-trajanovski (approved):
Seems good to me 👍

src/auth/EssUserAuthorization.ts
```diff
@@ -0,0 +1,99 @@
...
+    @inject(Tokens.ProposalDataSource)
+    protected proposalDataSource: ProposalDataSource,
+    @inject(Tokens.VisitDataSource) protected visitDataSource: VisitDataSource
+  ) {
+    super(userDataSource, sepDataSource, proposalDataSource, visitDataSource);
+  }
+
+  private decode(token: string) {
+    const PING_PUBLIC_CRT = process.env.PING_PUBLIC_CRT;
+    const PING_SIGNING_ALGORITHM = 'RS256';
```
mr-field, line 40: I know it's a micro-optimisation, but could this constant be declared in the class scope rather than the function's? That should avoid the need to re-initialise the variable every time the function is called (not sure if Node is smart enough to avoid it by itself)

mr-field (commented):
Looks good to me, with a really minor optimisation suggested.

I've noticed the code copies some data from the external identity system to the DB for every user, such as their name and email. This ensures you always have the data you need in the DB - for example a user can look up other users by name without triggering any queries to the external identity system. However, I was wondering if this could be a problem if users change their details in the external system? For example, if they changed their email but didn't log into the proposal app afterwards, am I right in thinking that the proposals DB would still have their old email? And if another user added them to a proposal at this point, would they know they've been added?

jekabs-karklins:
> Looks good to me, with a really minor optimisation suggested.
> 
> I've noticed the code copies some data from the external identity system to the DB for every user, such as their name and email. This ensures you always have the data you need in the DB - for example a user can look up other users by name without triggering any queries to the external identity system. However, I was wondering if this could be a problem if users change their details in the external system? For example, if they changed their email but didn't log into the proposal app afterwards, am I right in thinking that the proposals DB would still have their old email? And if another user added them to a proposal at this point, would they know they've been added?

You are exactly right. The mitigation we will add is to update user details every time when the user logs in.

PR #658, opened 2022-05-16
feat: Enable instrument management for STFC

mr-field:
## Description
**Please review the changes by commit**. Each commit has a description of what I was doing and why. They have been rebased to ensure they make logical sense.

The main point of this PR is to allow STFC instrument scientists to view all proposals for their facility only, without showing any proposals assigned to others.

Instead of adding a new "Facility" object to the app, STFC facilities are represented as Instruments. When running in STFC mode, the app expects the `ISIS`, `LSF`, `HPL` and `ARTEMIS` instruments to be present in the database. If it finds them, it automatically links STFC instrument scientists to these instruments when they log in, based on the STFC role. Scientists (or no-longer-scientists) will also be automatically removed from any instruments they don't manage anymore at the same time.

This PR also changes the behaviour of the instrument scientist proposal table in the STFC setup. It now shows all proposals submitted to calls which are assigned to an instrument assigned to the scientist. This removes the need for a user officer to assign proposals directly to an instrument before they can appear on the list, and it's the preferred way of showing instruments for STFC. The ESS confirmed they would like to keep the current behaviour for that list instead.

### After merging
After merging this PR, these instruments should be created in the STFC environement:
- ISIS
- LSF
- HPL
- ARTEMIS

We should also tell the User Office to link all calls to an instrument so that proposals submitted to them appear in the instrument scientist view.

## Motivation and Context
It allows STFC instrument scientists to view proposals for their facility, and automatically assigns them to the right facility.

## How Has This Been Tested
- Unit tests have been added to check the STFC auto-assign functionality.
- E2E tests: https://github.com/UserOfficeProject/user-office-frontend/pull/917
- I have also tested the changes manually.

## Fixes
Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/1


src/datasources/postgres/ProposalDataSource.ts
```diff
@@ -501,12 +501,22 @@ export default class PostgresProposalDataSource implements ProposalDataSource {
...
       .from('proposal_table_view')
-      .join('instruments', {
-        'instruments.instrument_id': 'proposal_table_view.instrument_id',
+      .join('instruments', function () {
+        this.on({
+          'instruments.instrument_id':
+            'proposal_table_view.proposal_instrument_id',
+        }).orOn({
+          'instruments.instrument_id': 'proposal_table_view.call_instrument_id',
+        });
```
mr-field, line 510: This change is making some of the E2E tests fail. We should check what the @UserOfficeProject/swap team thinks of this, then either update the test or move it back to the STFC datasource

mr-field: I've now moved the logic back to the STFC datasource. The ESS setup will keep the previous way of selecting proposals

PR #663, opened 2022-05-23
fix(auth): updating user details on login

jekabs-karklins:
## Description

This fix updates user details from OrcID upon every login.

## Motivation and Context

We want to keep the UOS database up to date with the changes from OrcID because data entered in OrcID is considered the source of truth. Users might change their name, surname, or email on OrcID.

## How Has This Been Tested

- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2613

## Changes

src/auth/EssUserAuthorization.ts

## Depends on

N/A


src/auth/EssUserAuthorization.ts
```diff
@@ -55,6 +55,13 @@ export class EssUserAuthorization extends UserAuthorization {
 
     const user = await this.userDataSource.getByOrcID(decoded.ORCID);
     if (user) {
+      await this.userDataSource.update({
+        ...user,
+        firstname: decoded.FirstName,
+        lastname: decoded.LastName,
+        email: decoded.mail,
```
martin-trajanovski, line 62: Do we need to perform a check for the email if it is a valid one or this is already trusted source that is sending only valid email addresses?

jekabs-karklins: It's from PING JWT so we can trust it

martin-trajanovski (approved):

PR #668, opened 2022-06-02
[WIP] feat: Call should end explicitly instead of endCall date

martin-trajanovski:
## Description

Call should become inactive when the cycle ends not on call end date. Added `is_active` flag on the call to be able to explicitly deactivate/activate calls.

## Motivation and Context

Previously we were making a call inactive on the call ended date.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2465

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/956


martin-trajanovski:
@bolmsten you can double check this now and confirm if the changes are correct.

bolmsten:
This looks good for ESS, can STFC verify that this would work for them as well?

src/datasources/postgres/CallDataSource.ts
```diff
@@ -48,21 +48,25 @@ export default class PostgresCallDataSource implements CallDataSource {
...
-      query
-        .where('start_call', '>=', currentDate)
-        .orWhere('end_call', '<=', currentDate);
+      query.where('is_active', false);
     }
 
+    /**
+     * NOTE: We are comparing dates instead of using the call_ended flag,
+     * because the flag is set once per hour and we could have a gap.
+     * TODO: Maybe there is a need to use the timezone setting here but not quite sure about it. Discussion is needed here!
```
simonfernandes, line 59: I would have thought this would be fine as `new Date().toISOString()` and start_call and end_call is all in UTC

simonfernandes (approved):

src/datasources/postgres/CallDataSource.ts
martin-trajanovski: @simonfernandes thank you I can remove the comment then. 🙂

src/datasources/postgres/CallDataSource.ts
```diff
@@ -48,21 +48,25 @@ export default class PostgresCallDataSource implements CallDataSource {
...
+     */
+    const currentDate = new Date().toISOString();
     if (filter?.isEnded === true) {
```
jekabs-karklins, line 62: Is the filter value `isEnded` used? If no service is using it we could consider removing it.

jekabs-karklins (approved):

PR #672, opened 2022-06-07
fix: Prevent random STFC logouts by reducing UOWS requests

mr-field:
## Description
Regenerates the UOWS client using the latest generator version. This includes changes which reduce calls made to the UOWS (from https://github.com/UserOfficeProject/user-office-lib/pull/131 and https://github.com/UserOfficeProject/user-office-lib/pull/134). The result of this should be much more reliable communication with the UOWS, reducing the problem of users being logged out when the UOWS receives too many requests. When testing locally, my setup has gone from reliably logging me out whenever I access the User Officer dashboard to always working on all pages I tested.

As part of this, Typescript was automatically updated to its latest version, because the `uows-client-generator` package now requires a later version of Typescript than the one previously used in this app. The ESlint plugins have been updated to match the new Typescript version.

The new version of TypeScript also spotted a problem with the SEP authorization check, which I fixed in 82879091a40698c8892e94a9ddbdae265f4367ab.

## Motivation and Context
The UOWS stops responding on time if it receives too many requests. The latest version of `uows-client-generator` mitigates the problem.

## How Has This Been Tested
By loading GraphQL-intensive frontend pages on my local setup.

## Fixes
Part of https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/580

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,4 @@ package-lock.json
 node_modules/
 .github/workflows/
 README.md
+/src/datasources/stfc/UOWSSoapInterface.ts
--- a/package-lock.json
+++ b/package-lock.json
@@ -77,7 +77,7 @@
         "@types/yup": "^0.29.13",
         "@typescript-eslint/eslint-plugin": "^5.7.0",
         "@typescript-eslint/parser": "^5.7.0",
-        "@user-office-software/uows_client_generator": "^2.3.0",
+        "@user-office-software/uows_client_generator": "^3.0.0",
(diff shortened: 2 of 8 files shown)


simonfernandes (approved):

ACLay (commented):
I realise 4.7.3 still meets the criteria of compatible with 4.5.4 in the package.json devdependencies, but since you mentioned needing to upgrade typescript, would it be worthwhile to update that reference?

ACLay (approved):

PR #678, opened 2022-06-08
fix: Fix permission issues with isMemberOfSEP where we should send sepId instead of userId

martin-trajanovski:

### DIFF
--- a/src/auth/UserAuthorization.ts
+++ b/src/auth/UserAuthorization.ts
@@ -70,7 +70,7 @@ export abstract class UserAuthorization {
     agent: UserWithRole | null,
     sepId?: number
   ): Promise<boolean> {
-    if (!agent || !agent.currentRole || !sepId) {
+    if (!agent || !agent.currentRole) {
       return false;
     }
 
@@ -81,8 +81,7 @@ export abstract class UserAuthorization {
 
   async listReadableUsers(
     agent: UserWithRole | null,
-    ids: number[],
-    sepId?: number
+    ids: number[]
   ): Promise<number[]> {
     if (!agent) {
       return [];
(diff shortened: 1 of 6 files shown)


mr-field (commented):
Is it necessary to pass `sepId` to the `UsersQuery`? It seems like every SEP member would have permission to get the full results, so maybe it would be easier to just make `sepId` optional in `isMemberOfSEP` and not pass any IDs when we define `isSEPMember`. If we do that, the datasource should check id the user is on any SEP without the need to pass a `sepId` to the query.

martin-trajanovski:
> Is it necessary to pass `sepId` to the `UsersQuery`? It seems like every SEP member would have permission to get the full results, so maybe it would be easier to just make `sepId` optional in `isMemberOfSEP` and not pass any IDs when we define `isSEPMember`. If we do that, the datasource should check if the user is on any SEP without the need to pass a `sepId` to the query.

Thanks @mr-field I just pushed the fix as you suggest and it should work.

mr-field (commented):
The change looks good, but I think something similar is also happening [here](https://github.com/UserOfficeProject/user-office-backend/blob/9bd37972e1aff11675380335592e021b7ce8c7d9/src/queries/SEPQueries.ts#L156). Unless I'm misundertanding the code, that check will also fail for the same reason (we're passing something which isn't a SEP ID). Maybe we can get the SEP ID of the proposal there and use it for the check?

martin-trajanovski:
> The change looks good, but I think something similar is also happening [here](https://github.com/UserOfficeProject/user-office-backend/blob/9bd37972e1aff11675380335592e021b7ce8c7d9/src/queries/SEPQueries.ts#L156). Unless I'm misundertanding the code, that check will also fail for the same reason (we're passing something which isn't a SEP ID). Maybe we can get the SEP ID of the proposal there and use it for the check?

@mr-field thanks for the tips. I will definitely take a look.

jekabs-karklins (approved):
Looks good

PR #680, opened 2022-06-09
feat: Add email to users and allow instrument scientist to see comments on proposal

bolmsten:
See:

https://github.com/UserOfficeProject/user-office-frontend/pull/951

### DIFF
--- a/src/auth/StfcUserAuthorization.spec.ts
+++ b/src/auth/StfcUserAuthorization.spec.ts
@@ -8,6 +8,7 @@ import { Instrument } from '../models/Instrument';
 import { StfcUserAuthorization } from './StfcUserAuthorization';
 
 jest.mock('../datasources/stfc/UOWSSoapInterface.ts');
+jest.mock('../utils/LRUCache');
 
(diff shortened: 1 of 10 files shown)


src/models/User.ts
```diff
@@ -89,6 +89,7 @@ export class BasicUserDetails {
     public organisation: string,
     public position: string,
     public created: Date,
-    public placeholder: boolean
+    public placeholder: boolean,
+    public email: string
```
martin-trajanovski, line 93: Exposing the email here in the BasicUserDetails might not be GDPR compliant. What do you think about that @bolmsten?

martin-trajanovski (commented):
Just a little concerned about the GDPR because I know this data for basic user details can be accessed by anyone I think. Even if you are just a normal user you can query that and get all other users emails.

src/models/User.ts
bolmsten: it is now restricted to only user officer and instrument scientist

jekabs-karklins (approved):
Looks good

PR #683, opened 2022-06-14
feat: Speed up STFC setup by caching UOWS results

mr-field:
## Description
The backend now caches most results from the UOWS. This includes validity of user tokens, person details, and roles. The data is stored in an LRU cache for a short time, so that any requests for the same data will not need a trip to the UOWS.

The `getUsers` function in the STFC user datasource has been left "uncached", as it's called very frequently as the user types surnames in a dialog box. I suspect that trying to cache results there would only lead to a large amount of useless results flooding the cache.

To help with evaluating the effectiveness of the caches, I've set them to log usage statistics every 5 minutes. This can be changed if people feel it would crowd the logs too much.

As part of this, I've implmented the cache myself. The main reason was that most of the caches I found didn't quite fit our use case in here, either because they couldn't auto-remove results or they needed separate instances of standalone cache DBs. I can use an existing one if people prefer. The cache I've implemented should be generic enough to be reused in other parts of the app, and it features 100% unit test coverage.

## Motivation and Context
This speeds up many GraphQL queries and mutations in the STFC setup, as it cuts down on the number of webservice calls required to execute them.

## How Has This Been Tested
- Unit tests
- Manually on my machine (the speedup looks noticeable when switching roles, as they now oad immediately)

## Fixes
Part of https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/580


TickleThePanda (commented):
From my understanding of the event loop, this is safe in JS and will be executed in a single loop. I can't see any events, timeouts, or promises that break the functions across loops.

mr-field:
> Since LRUCache seems to be reusable, would be nice to see LRU as an NPM package so it can be reused in other projects.

@jekabs-karklins I've opened an issue for moving this to its own NPM package: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/623

For now, I'd like to merge it here to prevent problems after we open up the STFC instance to more people, which will be very soon.

PR #685, opened 2022-06-20
fix: grant reading users for access token

jekabs-karklins:
## Description

This query
```
query {
  proposals(filter: { instrumentId: 4 }) {
    proposals {
      proposer {
        firstname
      }
    }
  }
}
```

Will fail if requested with the API access token with reading privileges

I believe this change is related to https://github.com/UserOfficeProject/user-office-backend/issues/678 , so please any feedback from @martin-trajanovski is appreciated because you looked into similar issue with SEP.


bolmsten (approved):

martin-trajanovski (commented):
@jekabs-karklins if this change is tested and it works I am fine with it. Because we control the `isApiAccessToken` anyway.

jekabs-karklins:
> @jekabs-karklins if this change is tested and it works I am fine with it. Because we control the `isApiAccessToken` anyway.

I thought so too

PR #691, opened 2022-06-29
feat: Add seps to a call

martin-trajanovski:
## Description

This PR is adding SEPs to a call. It creates new database table (`call_has_seps`) to connect SEPs to a call. Also added `seps` property to the update and create call mutation so when call is created/updated SEPs could be also saved on it.
 
## Motivation and Context

There was no call to SEPs connection and in the SEP tabs we were just showing all of the calls in the filters. Also now when assigning proposal to a SEP we filter only SEPs that are assigned to a proposals call.

## How Has This Been Tested

- manual testing
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2640

## Changes

https://user-images.githubusercontent.com/6333063/177553192-c8ac9451-1b24-4709-9528-3b2a1b038d88.mp4

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/960

### DIFF
--- a/src/datasources/postgres/CallDataSource.ts
+++ b/src/datasources/postgres/CallDataSource.ts
@@ -21,13 +21,7 @@ export default class PostgresCallDataSource implements CallDataSource {
       .del()
       .from('call')
       .returning('*')
-      .then((call: CallRecord[]) => {
-        if (call === undefined || call.length !== 1) {
-          throw new Error(`Could not delete call with id:${id}`);
-        }
-
-        return createCallObject(call[0]);
-      });
+      .then((call: CallRecord[]) => createCallObject(call[0]));
   }
   async getCall(id: number): Promise<Call | null> {
     return database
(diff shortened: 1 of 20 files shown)


src/datasources/postgres/CallDataSource.ts
```diff
@@ -112,7 +112,7 @@ export default class PostgresCallDataSource implements CallDataSource {
       .returning('*')
       .then((call: CallRecord[]) => {
         if (call.length !== 1) {
-          throw new Error('Could not create call');
+          return null;
```
jekabs-karklins, line 115: Why is this returning null? I believe that in most DataSources we throw an error, which is then handled in a central place and the rejection is returned.
See here https://confluence.esss.lu.se/display/SWAP/UO+Diagram - DB layer on mutations either executes the request of throws an error if unsuccessful.

src/datasources/postgres/CallDataSource.ts
```diff
@@ -233,6 +233,47 @@ export default class PostgresCallDataSource implements CallDataSource {
...
+    sepIds: number[];
+  }): Promise<boolean> {
+    const valuesToInsert = args.sepIds.map((sepId) => ({
+      sep_id: sepId,
+      call_id: args.callId,
+    }));
+
+    const sepsAssigned = await database.transaction(async (trx) => {
+      try {
+        /*
```
jekabs-karklins, line 247: Is this something that still needs to be done?

src/datasources/postgres/CallDataSource.ts
```diff
@@ -233,6 +233,47 @@ export default class PostgresCallDataSource implements CallDataSource {
...
+          .into('call_has_seps')
+          .transacting(trx);
+
+        return await trx.commit(sepsAssignedTransaction);
+      } catch (error) {
+        logger.logException(
+          `Could not assign seps to a call '${args.callId}'`,
+          error
+        );
+        trx.rollback();
```
jekabs-karklins, line 266: Maybe after rollback, it should throw another exception , so that rejection is returned

src/datasources/postgres/CallDataSource.ts
martin-trajanovski: Thanks I will doublecheck this.

src/datasources/postgres/CallDataSource.ts
martin-trajanovski: No I am just explaining in the comment what we are doing with the next transaction.

src/datasources/postgres/CallDataSource.ts
martin-trajanovski: I doublechecked and I will remove that throw from here completely because it is not needed. I have tested and it seems that if the execution goes into the then block call is created for sure. If there is an error while creating database error is thrown and we can catch that one. So basically there is no effect if you throw into the then block. Nothing happens.

src/datasources/postgres/CallDataSource.ts
martin-trajanovski: Changed the approach a little bit. Now everything is contained in the same database transaction and if there is some problem creating or updating call whole transaction is rolled back. This keeps the mutations more atomic and prevents from unexpected issues where we do all the actions in one transaction.

jekabs-karklins (approved):

PR #694, opened 2022-07-05
fix: Hotfix for STFC instrument assignment

simonfernandes:
This fixes instrment scientists (and others) not being assigned to instruments.

We were filtering out all of the non duplicate roles, instead of the inverse. On dev and for me at least, the "User Officer" role was the only duplicate, which is probably why we didn't notice.

Part of UserOfficeProject/user-office-project-issue-tracker#628

### DIFF
--- a/src/auth/StfcUserAuthorization.ts
+++ b/src/auth/StfcUserAuthorization.ts
@@ -210,7 +210,7 @@ export class StfcUserAuthorization extends UserAuthorization {
       // The UOWS sometimes returns duplicate roles. We remove them here
       const uniqueRoles = stfcRoles.filter(
         (role, index) =>
-          stfcRoles.findIndex((r) => r.name == role.name) !== index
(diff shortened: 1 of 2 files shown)


src/auth/StfcUserAuthorization.ts
```diff
@@ -210,7 +210,7 @@ export class StfcUserAuthorization extends UserAuthorization {
       // The UOWS sometimes returns duplicate roles. We remove them here
       const uniqueRoles = stfcRoles.filter(
```
TickleThePanda, line 211: Is there any way we can add unit tests for this?

TickleThePanda (changes requested):

mr-field (approved):

simonfernandes:
I put the wrong issue number in my last commit ☹

martin-trajanovski (approved):

TickleThePanda (approved):

PR #695, opened 2022-07-06
Mergeback after hotfix release 06/07/2022

simonfernandes:
## Description

Merge master into develop after hotfix


simonfernandes:
Needs to be done manually rather than through GitHub interface

PR #698, opened 2022-07-07
feat: Add predefined_messages table and expose CRUD endpoints

martin-trajanovski:
## Description

This PR is adding predefined_messages database table and CRUD operations around it. This table is used to store messages that will be reused in the comments and messages area throughout the app.

## Motivation and Context

Previously there was no mechanism to define and reuse messages and every time we needed to use the same message we need to copy paste or retype it again.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2599

## Changes

https://user-images.githubusercontent.com/6333063/179757655-56f28422-8232-4a23-bc81-571cbd93c058.mp4

## Depends on

Frontend PR: https://github.com/UserOfficeProject/user-office-frontend/pull/967
Validation PR: https://github.com/UserOfficeProject/user-office-lib/pull/143


jekabs-karklins:
Would it be possible to separate these two visual elements, by grouping them? Either by border or Paper or just more padding.
image

db_patches/0125_AddPredefinedMessagesTable.sql
```diff
@@ -0,0 +1,20 @@
+DO
+$DO$
+BEGIN
+  IF register_patch('AddPredefinedMessagesTable.sql', 'Martin Trajanovski', 'Add predefined_messages database table', '2022-07-07') THEN
+  BEGIN
+
+    CREATE TABLE predefined_messages (
+      predefined_message_id SERIAL PRIMARY KEY,
+      title TEXT,
+      key VARCHAR(20) NOT NULL,
```
jekabs-karklins, line 10: you could have `key` as a unique primary key

src/events/event.enum.ts
```diff
@@ -234,4 +237,16 @@ export const EventLabel = new Map<Event, string>([
     'Event occurs when the time slot booking is re-opened in the scheduler',
   ],
   [Event.INSTRUMENT_DELETED, 'Event occurs when the instrument is removed'],
+  [
```
jekabs-karklins, line 240: Are these events used somewhere or is this just futureproof

jekabs-karklins (approved):

ACLay (commented):
Could there be some comments added to some of the PredefinedMessage classes explaining what these messages are for? I read through an awful lot of this assuming it would be some sort of component for adding common headings or text to multiple questionnaire templates or pages, and it only made sense when I saw the video in the PR.

db_patches/0125_AddPredefinedMessagesTable.sql
martin-trajanovski: I am not sure if this should be unique because the `key` is used for another purpose. This `key` is just for filtering the messages by some specific key when we query them. For example if you have `key=user` you want to show only messages that are stored for users and not the ones for the managers.

ACLay (approved):

PR #701, opened 2022-07-12
fix: Fix STFC email search

simonfernandes:
Second attempt at fixing the STFC email search, since I missed out the optional chaining last time 🙃

The UOWS returns null when users aren't found which is why it's necessary. I've amended the tests to reflect that, and also added a few other elvis operators to avoid thrown errors in other places.

I've double checked it this time 😅


keiranjprice101 (approved):

ACLay:
> We seriously need to add types to the UOWS client

Looks like we've already got https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/51 for that

mr-field:
Yes, I meant later down the line. For now let's fix the immediate problem

PR #704, opened 2022-07-20
259 send email to engineer when callcreated

mutambaraf:
## Description

Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/259

## Motivation and Context

As a support engineer, I need to support users while they submit their proposals. Typically, I receive a vast amount of support requests in the days before a call ends. Therefore, I want to know in advance when calls are scheduled to end so that I can make arrangements for the larger than usual volume of requests.

For example, it might be helpful to get an email when an user officer creates a call, so that I can save the start and end date in a calendar. Ideally, I would also get a reminder email close to the call end date.

## How Has This Been Tested

Is coved by call mutations create call test

## Fixes

Fix:https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/259

## Changes
Files changed: stfcEmailHandler.ts

## Depends on

https://git.devs.facilities.rl.ac.uk/isisbusapps/BisAppSettings/pulls/19
https://github.com/isisbusapps/docker-orchestration/pull/277


src/eventHandlers/email/stfcEmailHandler.ts
```diff
@@ -111,6 +112,41 @@ export async function stfcEmailHandler(event: ApplicationEvent) {
 
       return;
     }
+
+    case Event.CALL_CREATED: {
+      if (event?.call) {
+        const templateID = 'isis-call-created-pi';
+        const noficicationEmailAdress =
+          (process.env && process.env.FBSEMAIL) ||
+          'FacilitiesBusinessSystem@stfc.ac.uk';
```
vyshnavi1605, line 121: I don't think we would like to hardcode the email here. Also is the env variable in the `docker-compose.yml`? If not this will need to be added [here](https://github.com/isisbusapps/docker-orchestration/blob/develop/docker-compose.yml#L254). Then it can read `const noficicationEmailAdress = process.env.FBSEMAIL `

src/eventHandlers/email/stfcEmailHandler.ts
```diff
@@ -173,3 +209,25 @@ const uoRapidSubmissionEmail = (
   },
   recipients: [{ address: uoAddress }],
 });
+
+const piCallCreationEmail = function createNotificationEmail<T>(
```
vyshnavi1605, line 213: The pi doesn't create the call, so maybe we can rename `piCallCreationEmail` to something like `callCreationEmail`?

src/eventHandlers/email/stfcEmailHandler.ts
```diff
@@ -173,3 +209,25 @@ const uoRapidSubmissionEmail = (
...
+  _notificationInput: T,
+  _templateID: string,
+  _notificationEmailAddress: string
```
vyshnavi1605, line 216: To keep the naming convention similar, we could perhaps leave takeout the `_` before the names?

mutambaraf: I am trying to differentiate parameter variable from local variable to avoid mistakenly mutation of original variable affecting other parts of the code.

src/eventHandlers/email/stfcEmailHandler.ts
mutambaraf: I will create the PR for the email variable. I thought it would be nice to have a fall back when it is not found, the alternative is to log an error .Will make the necessary changes.

src/eventHandlers/email/stfcEmailHandler.ts
mutambaraf: Noted

martin-trajanovski (approved):
From our end I think it looks good. I see the changes are only in the stfcEmailHandler and we are not that much affected in ESS by this.

src/eventHandlers/email/stfcEmailHandler.ts
vyshnavi1605: Oh okay, that makes sense

src/eventHandlers/email/stfcEmailHandler.ts
```diff
@@ -111,6 +112,42 @@ export async function stfcEmailHandler(event: ApplicationEvent) {
 
       return;
     }
+
+    case Event.CALL_CREATED: {
+      if (event?.call) {
+        const templateID = 'isis-call-created-pi';
+        if (process.env && process.env.FBSEMAIL) {
+          const noficicationEmailAdress = process.env && process.env.FBSEMAIL;
```
vyshnavi1605, line 120: In the docker-orchestration PR the variable is defined as `FBS_EMAIL`, I think here will need to changed to be the same to map to right variable?

src/eventHandlers/email/stfcEmailHandler.ts
```diff
@@ -111,6 +112,42 @@ export async function stfcEmailHandler(event: ApplicationEvent) {
 
       return;
     }
+
+    case Event.CALL_CREATED: {
+      if (event?.call) {
+        const templateID = 'isis-call-created-pi';
+        if (process.env && process.env.FBS_EMAIL) {
```
simonfernandes, line 119: We should probably log an error if `FBS_EMAIL` is not specified as it will indicate a problem in our setup that we can fix

src/eventHandlers/email/stfcEmailHandler.ts
```diff
@@ -111,6 +112,42 @@ export async function stfcEmailHandler(event: ApplicationEvent) {
...
+          const noficicationEmailAdress = process.env.FBS_EMAIL;
```
simonfernandes, line 120: Typo in "notification"

vyshnavi1605 (approved):

simonfernandes (approved):

PR #707, opened 2022-08-04
feat(auth): Implement openid connect authorizer

jekabs-karklins:
## Description

This PR aims to add support for authentification with external identity providers using OpenId connect.

## Motivation and Context

As we are integrating with the Ping Identity Access Management platform which supports OpenId Connect which is a simple identity layer on top of the OAuth 2.0 protocol, the implementation has been made generic, so that it is possible to use any other Identity Provider platform implementing the OpenId connect.

This PR contains also an example of Google Authorization as can be seen from the example.development.env

IMPORTANT:
In the development, environment authorization is performed with a locally run simple OAUTH server (for more information inspect updated docker.compose file)
after this PR is merged if the OAuth authorizer is used. then new .env values must be set

## How Has This Been Tested

- [x] Manual tests
- [x]  E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2688

## Changes

Implemented base class OpenIdConnectAuthorization
Added new Authorizers, PingAuthorization, and example GoogleAuthorization

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/977


src/resolvers/types/User.ts
```diff
@@ -64,11 +64,14 @@ export class User implements Partial<UserOrigin> {
...
+  @Field(() => String, { nullable: true })
+  public oidcSub: string | null;
 
-  @Field()
-  public refreshToken: string;
+  @Field(() => String, { nullable: true })
+  public oidcRefreshToken: string | null;
+
+  @Field(() => String, { nullable: true })
+  public oidcAccessToken: string | null;
```
martin-trajanovski, line 74: Is it possible to query the user Access token now directly on the user object? Not sure how safe this is but just wanted to check.

martin-trajanovski (approved):
In general it looks good. Just few small comments left.

src/resolvers/types/User.ts
jekabs-karklins: Yes, this should be removed. No need to expose it to the client.

ACLay (changes requested):

src/config/ess/configureESSEnvironment.ts
```diff
@@ -83,6 +84,30 @@ async function enableDefaultEssFeatures() {
...
+    let endSessionUrl;
+    try {
+      endSessionUrl = client.endSessionUrl(); // try obtaining the end session url the standard way
+    } catch (e) {
+      endSessionUrl =
+        (client.issuer.ping_end_session_endpoint as string) ?? '/'; // try using PING ping_end_session_endpoint
+    }
+
+    await db.updateSettings({
+      settingsId: SettingsId.EXTERNAL_AUTH_LOGIN_URL,
```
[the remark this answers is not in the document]
jekabs-karklins: I kept the same flow as with STFC external auth (see) `configureSTFCEnvironment`. I agree that User officer should not change these values, and they are set here merely for client access. Changing the way the EXTERNAL_AUTH_LOGIN_URL is distributed could be a separate issue, and as this PR is already very large I would not want to add on top of it this change.

docker-compose.e2e.yml
```diff
@@ -7,6 +7,7 @@ services:
       DATABASE_URL: postgres://duouser:duopassword@db:5432/duo
       secret: qMyLZALzs229ybdQXNyzYRdju7X784TH
       tokenLife: 7d
+      protocol: http
```
ACLay: It looks like the use of this variable was removed in e343fe0c1336566212adac5f135f874bea6ff05c, can it be removed from here and `example.development.env`

src/auth/StfcUserAuthorization.ts
```diff
@@ -226,15 +215,23 @@ export class StfcUserAuthorization extends UserAuthorization {
...
+        await client.logout(token).catch(() => {
+          logger.logWarn('Failed to log out user', { token });
 
-    return;
+          return rejection('Failed to log out user', { token });
+        });
+      } else {
+        return rejection('No external token found in JWT', { token });
+      }
+    } catch (error) {
```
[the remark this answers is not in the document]
jekabs-karklins: I do not know, the reason, but I kept it as it was before to not change any flow for STFC authorization. This code used to sit in UserMutations, but I moved it in as the flow was STFC specific https://github.com/UserOfficeProject/user-office-backend/pull/707/files/f58e0e2445719c459fa49c23b113e4d39dfa6a90#diff-2590831eb92b22869cbdd92250fb1d366c439053179fa0c558765fb1313a111aL521

example.development.env
```diff
@@ -28,10 +28,20 @@ ORCID_CLIENT_SECRET=secret
...
+# AUTH_DISCOVERY_URL=https://accounts.google.com/.well-known/openid-configuration
+# AUTH_CLIENT_ID=123456789-abcdef.apps.googleusercontent.com
+# AUTH_CLIENT_SECRET=ABCDEF-0123456789
+
 # Other
 NODE_ENV=development
 # DEPENDENCY_CONFIG=<e2e|ess|stfc|test>
-DATABASE_URL=postgres://duouser:duopassword@127.0.0.1:5432/duo
+DATABASE_URL=postgres://duouser:duopassword@localhost:5432/duo
+protocol=http
```
ACLay: I don't think this variable is used anywhere now, can it be removed?

src/auth/EmailAsExternalTokenAuthorization.ts
```diff
@@ -0,0 +1,22 @@
+import 'reflect-metadata';
+import { injectable } from 'tsyringe';
+
+import { User } from '../models/User';
+import { UserAuthorization } from './UserAuthorization';
+
+@injectable()
+export class EmailAsExternalTokenAuthorization extends UserAuthorization {
```
ACLay, line 8: Not a necessity, but is there some way we can have this throw an error if it's used outside of tests? Maybe make it require an "allow insecure auth for tests" environment variable? An approve all authenticator seems like a risky thing to potentially have running in production

src/config/ess/configureESSEnvironment.ts
```diff
@@ -80,9 +81,34 @@ async function enableDefaultEssFeatures() {
       FeatureId.USER_MANAGEMENT,
       FeatureId.VISIT_MANAGEMENT,
       FeatureId.SAMPLE_SAFETY,
+      FeatureId.OAUTH,
     ],
     true
   );
+  if (OpenIdClient.hasConfiguration()) {
```
ACLay: It looks like if the OpenID client is missing config the app will still continue. Is that the case, and if so, is that behaviour desired?

src/auth/mockups/UserAuthorization.ts
```diff
@@ -1,27 +1,12 @@
...
-    @inject(Tokens.UserDataSource) protected userDataSource: UserDataSource,
-    @inject(Tokens.SEPDataSource) protected sepDataSource: SEPDataSource,
-    @inject(Tokens.ProposalDataSource)
-    protected proposalDataSource: ProposalDataSource,
-    @inject(Tokens.VisitDataSource) protected visitDataSource: VisitDataSource
-  ) {
-    super(userDataSource, sepDataSource, proposalDataSource, visitDataSource);
-  }
-
   async externalTokenLogin(token: string): Promise<User | null> {
```
ACLay: Should this UserAuthorization implementation also have a redirectUrl parameter added?

src/queries/UserQueries.ts
```diff
@@ -1,25 +1,21 @@
 import { logger } from '@user-office-software/duo-logger';
-import * as bcrypt from 'bcryptjs';
 // TODO: Try to replace request-promise with axios. request-promise depends on request which is deprecated.
```
ACLay: Since we've removed the request-promise import, can this comment be removed too

ACLay (changes requested):

src/auth/mockups/UserAuthorization.ts
jekabs-karklins: Yes, that's true

src/auth/EmailAsExternalTokenAuthorization.ts
jekabs-karklins: I have added the check, but I want to remove this Authorizer latter, as I don't think it is elegant

example.development.env
jekabs-karklins: True

src/config/ess/configureESSEnvironment.ts
jekabs-karklins: Yes, that is desired behavior. But I now see that the better fit for this setup is within the authorizer itself. So if you want to use this Auth all no need to do is map it in dependency config and set environmental variables

src/queries/UserQueries.ts
jekabs-karklins: 👌🏻

PR #709, opened 2022-08-10
refactor(graphql): rename getPageContent to pageContent

jekabs-karklins:
## Description

Query endpoint naming should not be preceded with `get`, but should be simply the name of the noun that is queries (e.g. `pageContent` instead of `getPageContent`)

## Motivation and Context

## How Has This Been Tested

- [x] Manual tests
- [x] E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2730

## Changes

This PR renames endpoint
- getPageContent -> pageContent

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/984


src/resolvers/queries/GetPageContentQuery.ts
```diff
@@ -5,7 +5,7 @@ import { PageName } from '../../models/Page';
 @Resolver()
```
martin-trajanovski, line 5: Maybe the file name should be also changed to just `PageContentQuery.ts`.

martin-trajanovski (changes requested):

src/resolvers/queries/GetPageContentQuery.ts
jekabs-karklins: 👍🏻

martin-trajanovski (approved):

TickleThePanda (approved):

bolmsten (approved):

PR #712, opened 2022-08-11
Swap 2732 add add questionnaire queries

jekabs-karklins:
## Description

This PR adds 2 useful queries for 3rd party integration
- proposalById
- questionByNaturalKey

## Motivation and Context

## How Has This Been Tested

- [x] Manual tests
- [x] E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2732

## Changes

This PR changes endpoints
- proposalById
- questionByNaturalKey

## Depends on
N/A

### DIFF
--- a/src/resolvers/queries/QuestionByNaturalKeyQuery.ts
+++ b/src/resolvers/queries/QuestionByNaturalKeyQuery.ts
@@ -1,0 +1,19 @@
+import { Arg, Ctx, Query, Resolver } from 'type-graphql';
+
+import { ResolverContext } from '../../context';
+import { Question } from '../types/Question';
+
+@Resolver()
+export class QuestionByNaturalKeyQuery {
+  @Query(() => [Question])
+  questions(
+    @Arg('naturalKey', () => String)
+    naturalKey: string,
+    @Ctx() context: ResolverContext
+  ) {
+    return context.queries.template.getQuestionByNaturalKey(
+      context.user,
+      naturalKey
+    );
+  }
+}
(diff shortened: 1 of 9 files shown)


src/resolvers/queries/QuestionByNaturalKeyQuery.ts
```diff
@@ -0,0 +1,19 @@
+import { Arg, Ctx, Query, Resolver } from 'type-graphql';
+
+import { ResolverContext } from '../../context';
+import { Question } from '../types/Question';
+
+@Resolver()
+export class QuestionByNaturalKeyQuery {
+  @Query(() => [Question])
+  questions(
```
martin-trajanovski, line 9: I am not sure about the naming here. Why we use `questions` when the `getQuestionByNaturalKey` returns type `Question`. Probably the endpoint should be called something similar to the class name as we are doing in the other queries. Something like `questionByNaturalKey` and the return type should be just `Question` instead of `[Question]`. Maybe I am missing something but this seems most logical to me, please correct me if I am wrong.

martin-trajanovski (changes requested):
In general this looks nice but there is one change request that you should take a look into.

src/resolvers/queries/QuestionByNaturalKeyQuery.ts
jekabs-karklins: You are right, I overlooked that. Thanks.

martin-trajanovski (approved):

TickleThePanda (approved):

PR #716, opened 2022-08-12
feat: Add PDF templates 

simonfernandes:
## Description

This is a PR that simply creates a new PDF template type. User officers can go into a new PDF template section and use the same controls that exist for other template types (create, edit, delete, archive). There is a simple page with a single input for the user officer to put in template contents. 

Eventually, the PDF template can be attached to the call (not in this PR), and when sent to the factory will render the PDF using the template defined.

Behind the scenes, the PDF type follows how existing templates have been done, with a few small differences since PDF templates are non-questionary. In order to re-use a lot of the existing template structure in the backend and components in the frontend, there are a few conditionals in the backend where PDF templates are handled slightly differently, e.g. in their creation. In the frontend, I have tried to re-use existing template components where possible, but have had to create a new (and simplistic) PDF Editor for example. To make use of more of the existing template components would require more complex refactoring.

A few things don't yet work: templates can't be deleted, and a call count should display alongside each template, and the title and desription is not editable after creation. If this isn't yet polished enough for the ESS, I could hide this behind a feature flag for now.

## Motivation and Context

We would like to sometimes render a proposal PDF using a custom template.

## How Has This Been Tested

- Unit tests and e2e tests
- Checked template data is created in table(s)

## Fixes

Part of UserOfficeProject/stfc-user-office-project#613

## Changes

Mainly new additions for PDF type. A few tweaks to Template type and components to accommodate.

## Depends on

Frontend: https://github.com/UserOfficeProject/user-office-frontend/pull/988


mutambaraf (approved):

mutambaraf (commented):
Then can also work in the event context.queries.pdfTemplate.getPdfTemplates return an empty array  it can be handled in catch or finally with a call back.

src/mutations/PdfTemplateMutations.ts
```diff
@@ -0,0 +1,71 @@
...
+        agent,
+        args,
+      });
+    }
+
+    try {
+      return await this.pdfTemplateDataSource.create(
+        template.templateId,
+        args.templateData,
+        agent?.id || -1 // Agent will never be null with Authorized decorator
```
mr-field, line 40: Is the `||` here because Typescript thinks `agent` might be undefined? If so, it may be best to either change its declaration in the function signature to just `agent: UserWithRole` (without the `| null`), or do something like `(agent as UserWithRole).id` here. It avoids having to use a potentially unsafe value if it ever does end up being `null`.

The same goes for the `||` in `TemplateMutations`

src/datasources/postgres/records.ts
```diff
@@ -599,6 +601,16 @@ export interface QuantityRecord {
   readonly quantity_id: string;
 }
 
+export interface PdfTemplateRecord {
```
mr-field, line 604: If you declare this in a module called `knex/types/tables`, Typescript should automatically type everything you do in the datasource: https://knexjs.org/guide/#typescript

src/datasources/mockups/PdfTemplateDataSource.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.init();
+  }
+
+  async delete(pdfTemplateId: number): Promise<PdfTemplate> {
+    if (dummyPdfTemplate.pdfTemplateId !== pdfTemplateId) {
+      throw new Error(`Template with ID ${pdfTemplateId} does not exist`);
+    }
+
+    const copyOfTemplate = dummyPdfTemplateFactory(dummyPdfTemplate);
+    dummyPdfTemplate.pdfTemplateId = 999; // mocking deleting template with ID
```
mr-field, line 33: Should the ID here be `pdfTemplateId`? That's what would normally happen - the method returns an object with the ID that was deleted

src/resolvers/types/Template.ts
```diff
@@ -89,4 +90,21 @@ export class TemplateResolver {
...
+    @Ctx() context: ResolverContext
+  ): Promise<PdfTemplate | null> {
+    const templates = await context.queries.pdfTemplate.getPdfTemplates(
+      context.user,
+      {
+        filter: { templateIds: [template.templateId] },
+      }
+    );
+
+    return templates?.filter(
```
mr-field, line 106: Do we need the `filter` here? We should only have fetched the templates with the right ID already

src/mutations/PdfTemplateMutations.ts
mutambaraf: I think Typescript only enforce type checking at compile time to really guard against null agent must be checked inside the code block if it is not null .My thought destructing with assignment might be an option {id=-1}:UserWithRole

mr-field: I don't think we should ever set the ID as `-1`, because that's an invalid user number and would cause a DB error anyway. I'd prefer if we let JavaScript fail normally should `agent` ever be `null`. We can't really assign any safe value if that is the case, and leaving it as `null` would at least give us a helpful error message like "Could not insert because ID is null" when trying to create the  template. This error would tell us what's actually going on, while if we tried to artificially set the ID as -1 we'd get somthing unhelpful like "Invalid foreign key reference", which doesn't tell us what the real problem is

src/datasources/postgres/TemplateDataSource.ts
```diff
@@ -140,6 +140,8 @@ export default class PostgresTemplateDataSource implements TemplateDataSource {
   }
 
   async deleteTemplate(templateId: number): Promise<Template> {
+    await database('pdf_templates').delete().where({ template_id: templateId });
```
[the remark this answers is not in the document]
simonfernandes: I think I made the delete function in the PdfTemplateDataSource for completeness, so that there is a mutation that can delete a PDF template as well as its corresponding template. But it isn't used.

This line could instead call the PdfTemplateDataSource delete function, but at that point I would have to remove the deletion of the original template from that function. I did think about doing it that way. I can change it if you think it would be better.

src/resolvers/types/Template.ts
simonfernandes: You're exactly right, as the constraint means there can only be one return result

src/datasources/postgres/records.ts
simonfernandes: That's really nice to know, thanks

src/mutations/PdfTemplateMutations.ts
simonfernandes: I went for the latter here as I can now see how the same has been done in other places.

src/datasources/mockups/PdfTemplateDataSource.ts
simonfernandes: I copied this, but the value of `copyOfTemplate` is returned and is the object with the ID that was deleted (it's always 1 and only works with ID 1 passed into this function). Setting the `dummyPdfTemplate` to a 'non-existent' ID doesn't really do much in the test.

src/datasources/PdfTemplateDataSource.ts
```diff
@@ -0,0 +1,19 @@
...
+import { PdfTemplatesArgs } from '../resolvers/queries/PdfTemplatesQuery';
+
+export interface PdfTemplateDataSource {
+  delete(pdfTemplateId: number): Promise<PdfTemplate>;
+  clonePdfTemplate(
+    sourceTemplateId: number,
+    newTemplateId: number
+  ): Promise<PdfTemplate>;
+  updatePdfTemplate(args: UpdatePdfTemplateArgs): Promise<PdfTemplate>;
+  create(
```
jekabs-karklins, line 12: Maybe this should be createPdfTemplate for consistency, and could use CreatePdfTemplateArg object

src/mutations/TemplateMutations.spec.ts
```diff
@@ -57,6 +67,29 @@ test('An userofficer can create template', async () => {
...
+  const description = 'The description';
+  const groupId = TemplateGroupId.PDF_TEMPLATE;
+
+  const template = await mutations.createTemplate(dummyUserOfficerWithRole, {
+    groupId,
+    name,
+    description,
+  });
+
+  expect(template instanceof Template).toBe(true);
```
jekabs-karklins, line 87: This syntax seems to be a little more suitable in this case
`expect(template).toBeInstanceOf(Template)`

jekabs-karklins (approved):
Small comments but looks good

mr-field (commented):
Thanks, looks better now. I have one last question about linking Calls and PDf templates. The PR description says that Calls and PDF templates can't yet be linked, and indeed the `pdfTemplateId` field of `CreateCallMutation` and `UpdateCallMutation` is ignored. Could we link them in this PR? I think it would be possible to just add a line for the template ID [here](https://github.com/UserOfficeProject/user-office-backend/blob/530bc6571df1b46d3759b22af31bb79fcc00a8b7/src/datasources/postgres/CallDataSource.ts#L210) and [here](https://github.com/UserOfficeProject/user-office-backend/blob/530bc6571df1b46d3759b22af31bb79fcc00a8b7/src/datasources/postgres/CallDataSource.ts#L89). The DB should already take care of rejecting the transaction if the template ID doesn't exist in the `template_id` table, so we shouldn't need to do much more than that

TickleThePanda:
> Could we link them in this PR?

There's another PR lined up with those changes in them.

mr-field (approved):

PR #717, opened 2022-08-12
refactor(graphql): rename query id params to have exact naming

jekabs-karklins:
## Description

To keep consistent with other endpoints and avoid ambiguity, query parameters name containing ID should be explicitly named so that the name contains the name of the object whose id it represents, e.g. for querying user instead of calling parameter `id`, it should be `userId`.

## Motivation and Context

## How Has This Been Tested

- [x] Manual tests
- [x] E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2731

## Changes

This PR changes endpoints
- user(userId: Int!): User
- basicUserDetails(userId: Int!): BasicUserDetails
- call(callId: Int!): Call
- getPageContent(pageId: PageName!): String
- proposalStatus(proposalStatusId: Int!): ProposalStatus
- proposalWorkflow(proposalWorkflowId: Int!): ProposalWorkflow

## Depends on
https://github.com/UserOfficeProject/user-office-frontend/pull/989


martin-trajanovski (changes requested):
This seems like a nice refactor to follow some general pattern throughout the application. The only concern that I have is that this will break the scheduler app because it is referencing the `User` and `BasicUserDetails` resolvers by `id` field. That should be changed now to reference by `userId` field instead. Let me know if you need some help with that but it should be pretty straight forward change. I will approve this after you create a PR in the scheduler with the needed changes.

jekabs-karklins:
> This seems like a nice refactor to follow some general pattern throughout the application. The only concern that I have is that this will break the scheduler app because it is referencing the `User`, `BasicUserDetails` and `Call` resolvers by `id` field. That should be changed now to reference by `userId`/`callId` field instead. Let me know if you need some help with that but it should be pretty straight forward change. I will approve this after you create a PR in the scheduler with the needed changes.

I had a similar concern and tested it, but there is actually no problem because the noun User still has the field "id". It's only the query param name that changes.

martin-trajanovski:
> > This seems like a nice refactor to follow some general pattern throughout the application. The only concern that I have is that this will break the scheduler app because it is referencing the `User`, `BasicUserDetails` and `Call` resolvers by `id` field. That should be changed now to reference by `userId`/`callId` field instead. Let me know if you need some help with that but it should be pretty straight forward change. I will approve this after you create a PR in the scheduler with the needed changes.
> 
> I had a similar concern and tested it, but there is actually no problem because the noun User still has the field "id". It's only the query param name that changes.

OK I understand. It is good that you have tested it.

martin-trajanovski (approved):

keiranjprice101 (approved):

PR #718, opened 2022-08-17
feat: add logging on error

TickleThePanda:
Refs https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/645
Refs https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/18

## Description

Adds an error logging plugin which puts any GraphQL errors to the logging framework. Using `info` as an error over GraphQL might not actually be an error.

## Motivation and Context

Currently looking at load testing our services and they don't let the server know when they've failed, if there's been failed requests. In addition, this means we're not relying on the client providing information to the server.

## How Has This Been Tested

Running the application locally and E2E using https://github.com/UserOfficeProject/user-office-frontend/pull/993.

## Fixes

N/A - just part of other issues

## Changes

Creates a simple plugin and adds it to the `src/middlewares/graphql.ts`.

## Depends on

N/A

### DIFF
--- a/src/middlewares/graphql.ts
+++ b/src/middlewares/graphql.ts
@@ -1,3 +1,4 @@
+import { logger } from '@user-office-software/duo-logger';
 import {
   ApolloServerPluginInlineTraceDisabled,
   ApolloServerPluginLandingPageDisabled,
@@ -5,6 +6,10 @@ import {
   ApolloServerPluginUsageReporting,
 } from 'apollo-server-core';
 import { ApolloServer } from 'apollo-server-express';
+import type {
+  ApolloServerPlugin,
+  BaseContext,
+} from 'apollo-server-plugin-base';
 import { Express } from 'express';
 import { GraphQLError } from 'graphql';
 import { container } from 'tsyringe';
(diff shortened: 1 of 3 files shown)


src/middlewares/graphql.ts
```diff
@@ -42,6 +47,18 @@ const apolloServer = async (app: Express) => {
     }
   );
 
+  const errorLoggingPlugin: ApolloServerPlugin<BaseContext> = {
+    async requestDidStart() {
+      return {
+        async willSendResponse({ response: { errors } }) {
```
mr-field, line 53: Is this better than using the [`didEncounterErrors`](https://www.apollographql.com/docs/apollo-server/integrations/plugins-event-reference/#didencountererrors) event? Looking at the docs, it seems to be made for this sort of use case

TickleThePanda: Ah, initially I read only `parsing` in that. Sounds like a better event!

mr-field (approved):

src/middlewares/graphql.ts
```diff
@@ -42,6 +47,18 @@ const apolloServer = async (app: Express) => {
     }
   );
 
+  const errorLoggingPlugin: ApolloServerPlugin<BaseContext> = {
```
jekabs-karklins, line 50: This could go into a separate file, but I trust your judgment of the best place for it.

jekabs-karklins (approved):

src/middlewares/graphql.ts
TickleThePanda: Thanks for the feedback.

I did consider it but I thought it didn't take away from the usability or understanding of this file (given its low complexity, coupling with the server implementation, the low size of this change/file, and how we've already set up the remaining things in here).

I'll go ahead and merge now.

PR #727, opened 2022-09-07
Use flat directory structure for emails

mr-field:
## Description
The backend now expects a flat directory structure in the emails directory, where emails are stored in a format like
`emails/<email name>.<file type>.pug`.

## Motivation and Context

This allows them to be mounted in using a Kubernetes ConfigMap, which doesn't support nested directories.

## Fixes
Part of https://github.com/isisbusapps/docker-orchestration/issues/148

### DIFF
--- a/src/eventHandlers/MailService/SMTPMailService.ts
+++ b/src/eventHandlers/MailService/SMTPMailService.ts
@@ -47,6 +47,12 @@ export class SMTPMailService extends MailService {
           relativeTo: path.resolve(process.env.EMAIL_TEMPLATE_PATH || ''),
         },
       },
+      getPath: (type, template) => {
+        return path.join(
+          process.env.EMAIL_TEMPLATE_PATH || '',
+          `${template}.${type}`
+        );
+      },
     });
   }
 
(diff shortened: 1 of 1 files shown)


ACLay (commented):
Are the ESS going to need to make changes to their templates too? Or is the SMTPMailService one of the things that only we use?

mr-field:
> Are the ESS going to need to make changes to their templates too? Or is the SMTPMailService one of the things that only we use?

We're the only ones sending emails with SMTP, the ESS uses [`SparkPostMailService`](https://github.com/UserOfficeProject/user-office-backend/blob/develop/src/eventHandlers/MailService/SparkPostMailService.ts) instead (injected from their [config](https://github.com/UserOfficeProject/user-office-backend/blob/b561245e4a8f814b05331f19199831941092df7b/src/config/dependencyConfigESS.ts#L74)).

src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -65,18 +71,17 @@ export class SMTPMailService extends MailService {
       sendMailResults.id = 'test';
     }
 
-    options.content.template_id = path.join(
+    const template = path.join(
```
ACLay: There's some similarities, should this be something that calls the new `getPath` method added above?

simonfernandes (approved):

src/eventHandlers/MailService/SMTPMailService.ts
```diff
@@ -71,10 +73,8 @@ export class SMTPMailService extends MailService {
       sendMailResults.id = 'test';
     }
 
-    const template = path.join(
-      process.env.EMAIL_TEMPLATE_PATH || '',
-      `${options.content.template_id}.html.pug`
-    );
+    const template =
+      this.getEmailTemplatePath('html', options.content.template_id) + '.pug';
```
ACLay, line 77: Rather than add the pug on the end, would it be better as this: `this.getEmailTemplatePath('html.pug', options.content.template_id);`

mr-field: I'd prefer to keep it this way, because it mirrors what `email-templates` does internally: `type` is treated like an enum and can only have 3 values or so. Also, the `type` doesn't necessarily need to be at the end of the path/filename but the extension does. If we keep it like this, it's more resilient should we ever change to a format like `type.emailName.pug`

ACLay (approved):

